PR #7, opened 2023-01-07
add rubygem-sqlite3 in install-dep.sh 

tjyang:
adding rubyrem-sqlit3 so that  ruby ruby-sqlite.rb example will work in Fedora


penberg:
Thank you @tjyang!

tjyang:
@penberg , I am learning Rust and Database stuff. Thank you for all your responses.

PR #9, opened 2023-01-08
server: HTTP listener

### LINKED ISSUE #8: SQL over HTTP
Serverless JavaScript apps, for example, integrate best with HTTP-based things (using the `fetch()` API). Let's add a SQL over HTTP protocol to `sqld`, which we can use to implement #4. We should support both single statements and transactional batches as PlanetScale's JavaScript driver does: https://github.com/planetscale/database-js

penberg:
Fixes #8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2211,6 +2211,7 @@ dependencies = [
  "crossbeam",
  "futures",
  "hex",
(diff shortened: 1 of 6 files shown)


penberg:
Obsoleted by #12

PR #12, opened 2023-01-09
Add support for queries over HTTP

MarinPostma:
This PR adds support for queries over HTTP.

It is enabled by runnign the server with the `--http-listen-addr <SOCKET_ADDR>` option.

The server is responding to `POST /` with a payload json with the following shape:
```json
{
        "statements": ["some SQL statement", "some other statement", ...]
    }
```

if the query should return rows, they are returned to the JSON format.

The BLOB rows are encoded in base64 with the standard alphabet and no padding.

unlike stateful connections that open a DB connection for every session, the HTTP approach uses a connection pool instead. The number of connections in the pool can scale up and down depending on load

I also added end variables for all cli args, to simplify configuration from docker-compose. Env vars are documented in the `--help` command for the binary.

### DIFF
--- a/server/src/http/mod.rs
+++ b/server/src/http/mod.rs
@@ -8,7 +8,7 @@ use bytes::{BufMut, Bytes, BytesMut};
 use hyper::body::to_bytes;
 use hyper::server::conn::AddrStream;
 use hyper::service::make_service_fn;
-use hyper::{Body, Request, Response};
+use hyper::{Body, Method, Request, Response};
 use serde::Deserialize;
 use serde_json::Number;
 use tokio::sync::{mpsc, oneshot};
@@ -55,13 +55,14 @@ fn query_response_to_json(rows: QueryResponse) -> anyhow::Result<Bytes> {
     Ok(buffer.into_inner().freeze())
 }
 
-async fn handle_request(
+async fn handle_query(
     mut req: Request<Body>,
     sender: mpsc::Sender<(oneshot::Sender<Result<QueryResponse, BoxError>>, Query)>,
 ) -> anyhow::Result<Response<Body>> {
     let bytes = to_bytes(req.body_mut()).await?;
     let req: HttpQueryRequest = serde_json::from_slice(&bytes)?;
     let (s, resp) = oneshot::channel();
+    // TODO: send query batch instead
     let _ = sender
         .send((s, Query::SimpleQuery(req.statements.join(";"), Vec::new())))
         .await;
@@ -73,6 +74,16 @@ async fn handle_request(
             Ok(Response::new(Body::from(json)))
         }
         Err(_) => todo!(),
+    }
+}
+
+async fn handle_request(
+    req: Request<Body>,
+    sender: mpsc::Sender<(oneshot::Sender<Result<QueryResponse, BoxError>>, Query)>,
+) -> anyhow::Result<Response<Body>> {
+    match (req.method(), req.uri().path()) {
+        (&Method::POST, "/") => handle_query(req, sender).await,
+        _ => Ok(Response::builder().status(404).body(Body::empty()).unwrap()),
     }
 }
 
--- a/server/src/http/mod.rs
+++ b/server/src/http/mod.rs
@@ -33,7 +33,8 @@ fn query_response_to_json(rows: QueryResponse) -> anyhow::Result<Bytes> {
                         query::Value::Null => serde_json::Value::Null,
                         query::Value::Integer(i) => serde_json::Value::Number(Number::from(i)),
                         query::Value::Real(x) => serde_json::Value::Number(
-                            Number::from_f64(x).ok_or(anyhow::anyhow!("invalid float value"))?,
+                            Number::from_f64(x)
+                                .ok_or_else(|| anyhow::anyhow!("invalid float value"))?,
                         ),
                         query::Value::Text(s) => serde_json::Value::String(s),
                         query::Value::Blob(v) => {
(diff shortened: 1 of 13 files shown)


petehunt:
What does the response look like?

Additionally, I think it would be idea if I could provide placeholders in the SQL query. I'm thinking something like this in the request body:

```
{
  "query": "select email from users where user_id=:userId",
  "variables": {
    "userId": "1234"
  }
}
```

That way you wouldn't have to implement SQL escaping logic in your client if you don't want to.

Also - how would transactions work with this approach?

MarinPostma:
Hey @petehunt !

The API is still very crude, the response is just an array of JSON objects representing the rows.

Taking inspiration from PlanetScale serverless driver, I expect parameter binding and transaction management to be handled at the driver level. This makes sense since the HTTP API is stateless, and interactive transactions are disallowed

penberg:
@petehunt To add what @MarinPostma, the array of statements are run in a transaction similar to https://github.com/planetscale/database-js#transactions

penberg:
@petehunt here's an example JSON output:

```console
[penberg@turing ~]$ curl -d '{"statements": ["SELECT * FROM users"]}' localhost:8080
[{"email":"penberg@iki.fi"}] 
```

petehunt:
Sweet! I think that is a great compromise for transactions.

I still think parameter binding would be pretty useful though. If you had that your DB driver implementation would be trivial and much easier to bring to every platform under the sun. I also like curl-ability for adhoc queries :)

penberg:
@MarinPostma I think Pete's binding idea makes sense. We can work on that as a follow-up to this, of course.

MarinPostma:
I'm not against it, we need to think a bit about the design

haaawk:
I support the binding idea. Mostly because I think we should consider adding support for prepared statements and then binding parameters will be necessary.

server/src/http/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+                Ok::<_, Infallible>(service_fn(move |req| handle_request(req, sender.clone())))
+            }
+        }));
+
+    tokio::spawn(async move {
+        let mut pool = pool::Builder::new().build(db_factory, ());
+        while let Some((resp, query)) = receiver.recv().await {
+            if let Err(e) = poll_fn(|c| pool.poll_ready(c)).await {
+                tracing::error!("Connection pool error: {e}");
+                continue;
```
haaawk, line 106: Is it ok to drop the query here when the connection pool fails?
I'm not sure whether dropping a query and keeping the operation is the right choice here.
We have 2 other options:
1. fail the server
2. retry fetching the connection while keeping the same query

If dropping the query is what we really want then maybe we should at least include the query into the error message so that we know what was dropped?

server/src/http/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+use tokio::sync::{mpsc, oneshot};
+use tower::balance::pool;
+use tower::load::Load;
+use tower::{service_fn, BoxError, MakeService, Service};
+
+use crate::query::{self, Query, QueryError, QueryResponse, ResultSet};
+
+fn query_response_to_json(rows: QueryResponse) -> anyhow::Result<Bytes> {
+    let QueryResponse::ResultSet(ResultSet { columns, rows }) = rows;
+    let mut values = Vec::new();
```
haaawk, line 23: Nit: `s/Vec::new()/Vec::with_capacity(rows.len())/`?

server/src/http/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+                        query::Value::Null => serde_json::Value::Null,
+                        query::Value::Integer(i) => serde_json::Value::Number(Number::from(i)),
+                        query::Value::Real(x) => serde_json::Value::Number(
+                            Number::from_f64(x).ok_or(anyhow::anyhow!("invalid float value"))?,
+                        ),
+                        query::Value::Text(s) => serde_json::Value::String(s),
+                        query::Value::Blob(v) => {
+                            serde_json::Value::String(BASE64_STANDARD_NO_PAD.encode(v))
+                        }
+                    };
```
haaawk, line 42: Nit: would be nice to have a separate function for this conversion.

haaawk:
LGTM

server/src/http/mod.rs
MarinPostma: no it's not, but I fixed it upstream. Not RFR :)

MarinPostma: BTW, what's happening here is that the query is not executed, and an internal error will be returned to the user because the response channel was dropped

haaawk: Sounds good then. What's RFR btw?

MarinPostma: Ready for review

server/src/lib.rs
```diff
@@ -30,50 +36,87 @@ pub enum Backend {
...
+async fn run_service<S>(service: S, config: Config) -> Result<()>
+where
+    S: Service<(), Error = anyhow::Error> + Sync + Send + 'static + Clone,
+    S::Response: Service<Query, Response = QueryResponse, Error = QueryError> + Sync + Send,
+    S::Future: Send + Sync,
+    <S::Response as Service<Query>>::Future: Send,
+    // F: MakeService<(), Query, MakeError = anyhow::Error> + Sync + Send,
+    // F::Future: 'static + Send + Sync,
+    // F::Service: Service<Query, Response = QueryResponse, Error = QueryError> + Sync + Send,
+    // <F::Service as Service<Query>>::Future: Send,
```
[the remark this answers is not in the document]
MarinPostma: shoot, forgot to push

psarna (approved):

PR #32, opened 2023-01-13
libsql-client: Support local libSQL files

### LINKED ISSUE #30: client developer experience would be better if we could pass a local file from the start
I am testing this configuration:

```
  const config = {
    url: process.env.DATABASE_URL
  };
  const db = connect(config);
```

If `DATABASE_URL` is empty it works fine but it seems to be in memory.
If I pass a filename, I get:

```
	throw new TypeError('Only absolute URLs are supported');
		      ^
```

And if I try a `sqlite://` string, I get:

```
/Users/glaubercosta/sqldblog/node_modules/node-fetch/lib/index.js:1331
		throw new TypeError('Only HTTP(S) protocols are supported');
```

This forces the user to set up a local sqld for dispatch, but for usage in integration tests, for example, this is less optimal.
If a local file is passed, this should be a no-setup local sqlite instance.

penberg:
Fixes #30

### DIFF
--- a/packages/libsql-client/README.md
+++ b/packages/libsql-client/README.md
@@ -15,12 +15,14 @@ const rs = await db.execute("SELECT * FROM users");
 console.log(rs);
 ```
 
-You can also just run against local SQLite by dropping the `url` option from configuration:
+You can also just run against local SQLite with:
 
 ```typescript
 import { connect } from "@libsql/client"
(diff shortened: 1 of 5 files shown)


psarna (approved):

penberg:
bors merge

PR #35, opened 2023-01-13
change postgres' default port

glommer:
The postgres default port is 5432, so I suggest we default to the same, so that clients don't have to specify a port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ cargo run
 and connect to it with psql:
 
 ```console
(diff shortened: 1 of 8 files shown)


penberg:
@glommer Grep for 5000 and you'll find some more places where you need to change this.

glommer:
@penberg I changed a bunch of occurrences, hopefully all

penberg:
bors merge

PR #45, opened 2023-01-17
Full support for query parameters

MarinPostma:
This PR introduces full support for query parameters. Both positional and named parameters are supported.

The supported syntax is the same as the one described in https://www.sqlite.org/c3ref/bind_blob.html.
Unbound parameters are interpreted as NULL.

## HTTP query parameters

Parameters can also be bound in http request. The syntax is quite flexible:

* Request without params:
```json
{
    "statements": ["select * from users where name = 'adhoc'"]
}
```
or (syntaxes can be mixed in the same array):
```json
{
    "statements": [{"q": "select * from users where name = 'adhoc'"}]
}
```

* Request with params:
- positional:
```json
{
    "statements": [
        {"q": "select * from users where name = ?", "params": ["adhoc"]},
        {"q": "select * from users where name = ?1", "params": ["adhoc"]},
        {"q": "select * from users where name = $1", "params": ["adhoc"]}
    ]
}
```

- named:
```json
{
    "statements": [
        {"q": "select * from users where name = $name", "params": {"name": "adhoc"}},
        {"q": "select * from users where name = :name", "params": {"name": "adhoc"}},
        {"q": "select * from users where name = @name", "params": {"name": "adhoc"}},
        {"q": "select * from users where name = $1", "params": {"name": "adhoc"}}, # object is order sensitive
        {"q": "select * from users where name = ?", "params": {"name": "adhoc"}},
    ]
}
```

### Handling of BLOB

BLOBS are handled as base64 encoded string (standards alphabet, no padding), and are nested into an object for disambiguition:
```json
{
    "statements": [
    {"q": "select * from users where name = $name", "params": {"name": {"blob": "984HG3e"}}},
                                                                  # some b64 blob --^
    ]
}
```

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -27,7 +27,7 @@ pin-project-lite = "0.2.9"
 postgres-protocol = "0.6.4"
 prost = "0.11.3"
 regex = "1.7.0"
-rusqlite = { version = "0.28.0", features = [ "buildtime_bindgen", "column_decltype" ] }
+rusqlite = { version = "0.28.0", features = [ "buildtime_bindgen", "column_decltype", "modern_sqlite" ] }
 serde = { version = "1.0.149", features = ["derive"] }
 serde_json = "1.0.91"
 smallvec = "1.10.0"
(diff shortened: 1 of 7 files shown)


MarinPostma:
bors try

penberg:
@MarinPostma Do the positional ones work with SQLite? If not, we probably shouldn't do it because we want to preserve the ability to use the JavaScript driver, for example, with both SQLite (local development) and HTTP (production).

penberg:
bors merge

PR #47, opened 2023-01-18
docker-entrypoint.sh: Make listen addresses configurable

penberg:
Let's make the listen addresses configurable to give more flexibility at deployment time. For example, you might want to expose the PostgreSQL and HTTP services to public internet via IPv4, but use a cluster-private network for gRPC.


penberg:
bors merge

PR #52, opened 2023-01-18
Add `--http-auth <AUTH>` command line option

### LINKED ISSUE #49: HTTP basic authentication support
Add two new command line options, `--http-username` and `--http-password`, to enable HTTP basic authentication in the `sqld` HTTP server for basic access control.

penberg:
This adds a `--http-auth <AUTH>` command line option to `sqld`. If it is enabled, HTTP requests are denied unless the authornization string matches. You can use the mechanism with basic authentication or OAuth bearer tokens.

For example, to configure basic authentication, run:

```console
cargo run -- --http-listen-addr 127.0.0.1:8080 --http-auth "Basic cGVuYmVyZzpwZW5iZXJn"
```

and then access the URL with:

```console
curl -i http://penberg:penberg@127.0.0.1:8080
```

Fixes #49

### DIFF
--- a/sqld/src/http/mod.rs
+++ b/sqld/src/http/mod.rs
@@ -1,6 +1,8 @@
+pub mod auth;
 mod types;
 
 use std::future::poll_fn;
+use std::sync::Arc;
 use std::{convert::Infallible, net::SocketAddr};
 
 use base64::prelude::BASE64_STANDARD_NO_PAD;
@@ -21,6 +23,7 @@ use crate::http::types::HttpQuery;
 use crate::query::{self, Params, Queries, Query, QueryResponse, QueryResult, ResultSet};
 use crate::query_analysis::{final_state, State, Statement};
 
+use self::auth::Authorizer;
 use self::types::QueryObject;
 
 impl TryFrom<query::Value> for serde_json::Value {
@@ -159,10 +162,19 @@ async fn show_console() -> anyhow::Result<Response<Body>> {
 }
 
 async fn handle_request(
+    authorizer: Arc<dyn Authorizer + Send + Sync>,
     req: Request<Body>,
     sender: mpsc::Sender<Message>,
     enable_console: bool,
 ) -> anyhow::Result<Response<Body>> {
+    {
+        if !authorizer.is_authorized(&req) {
+            return Ok(Response::builder()
+                .status(hyper::StatusCode::UNAUTHORIZED)
+                .body(Body::empty())
+                .unwrap());
+        }
+    }
     match (req.method(), req.uri().path()) {
         (&Method::POST, "/") => handle_query(req, sender).await,
         (&Method::GET, "/console") if enable_console => show_console().await,
@@ -172,6 +184,7 @@ async fn handle_request(
 
 pub async fn run_http<F>(
     addr: SocketAddr,
+    authorizer: Arc<dyn Authorizer + Send + Sync>,
     db_factory: F,
     enable_console: bool,
 ) -> anyhow::Result<()>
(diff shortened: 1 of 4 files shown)


server/src/http/mod.rs
```diff
@@ -142,10 +142,32 @@ async fn show_console() -> anyhow::Result<Response<Body>> {
...
+                }
+            }
+            _ => {
+                return Ok(Response::builder()
+                    .status(hyper::StatusCode::UNAUTHORIZED)
+                    .body(Body::empty())
+                    .unwrap())
+            }
+        }
+    }
```
MarinPostma, line 170: can you factorize that?

server/src/http/mod.rs
```diff
@@ -142,10 +142,32 @@ async fn show_console() -> anyhow::Result<Response<Body>> {
...
+    expected_auth: Option<String>,
     req: Request<Body>,
     sender: mpsc::Sender<Message>,
     enable_console: bool,
 ) -> anyhow::Result<Response<Body>> {
+    if let Some(expected_auth) = expected_auth {
+        let headers = req.headers();
+        match headers.get(hyper::header::AUTHORIZATION) {
+            Some(actual_auth) => {
+                let actual_auth = actual_auth.to_str()?;
```
MarinPostma, line 154: we don't want to return an error here, but a response instead (bad request). We need better error handling, anyhow is terrible for our use case.

MarinPostma (changes requested):

server/src/http/auth.rs
```diff
@@ -0,0 +1,72 @@
...
+                    .map(|actual_auth| actual_auth == expected_auth)
+                    .unwrap_or(false)
+            } else {
+                false
+            }
+        }
+        None => true,
+    }
+}
+*/
```
MarinPostma, line 72: do we need that?

server/src/http/mod.rs
```diff
@@ -159,10 +163,20 @@ async fn show_console() -> anyhow::Result<Response<Body>> {
 }
 
 async fn handle_request(
+    authorizer: Arc<Mutex<Box<dyn Authorizer + Send + Sync>>>,
```
MarinPostma, line 166: why do we need a mutex here? we don't seem to need exclusive access anywhere

server/src/http/mod.rs
```diff
@@ -172,6 +186,7 @@ async fn handle_request(
 
 pub async fn run_http<F>(
     addr: SocketAddr,
+    authorizer: Arc<Mutex<Box<dyn Authorizer + Send + Sync>>>,
```
MarinPostma, line 189: looks like an `Arc<dyn Authorizer + Send + Sync>` would do the job here

MarinPostma (changes requested):

server/src/http/mod.rs
penberg: fixed

server/src/http/mod.rs
penberg: fixed

sqld/src/http/mod.rs
```diff
@@ -159,10 +162,19 @@ async fn show_console() -> anyhow::Result<Response<Body>> {
 }
 
 async fn handle_request(
+    authorizer: Arc<Box<dyn Authorizer + Send + Sync>>,
```
MarinPostma, line 165: `Box` is redundant: `Arc<dyn Authorizer + Send + Sync>` is sufficient

MarinPostma (changes requested):

sqld/src/http/mod.rs
penberg: fixed

MarinPostma (approved):
bors merge

PR #61, opened 2023-01-19
Use TLS for gRPC

honzasp:
Adds support for TLS on the gRPC connection between master and replicas. We use "mutual TLS" to authenticate both peers (client and server) with each other, so both peers need three files:

- CA certificate (root certificate): we connect only to peers with certificates that are signed by this certificate. Every cluster should have its own root certificate.
- Own certificate: the certificate that we show to the peer when authenticating.
- Own private key: the private key that corresponds to the certificate, we prove that we know the private key to the peer to authenticate ourselves to the peer.

The TLS support is optional and behind a command-line argument, plaintext gRPC is still the default.

I also included a small Python script to generate the server and client certificates for development.

Builds on Pekka's work: #42


MarinPostma (approved):
bors merge

PR #62, opened 2023-01-19
disable postgres

MarinPostma:
disable postgres protocol by default, and make HTTP the default instead (listen on 127.0.0.1:8080).

fix #33


penberg:
bors merge

PR #74, opened 2023-01-21
docker: Fix SQLD_GRPC_LISTEN_ADDR for replicas

penberg:
Let's *not* export SQLD_GRPC_LISTEN_ADDR by default; otherwise replica `sqld` picks it up via the CLI environment variables...


penberg:
bors merge

PR #78, opened 2023-01-21
Pass along top-level error messages

CodingDoug:
Handles cases where there are no result sets (such as SQL syntax errors)


penberg:
bors merge

PR #88, opened 2023-01-23
improve syntax error message

MarinPostma:
improve the error message when a syntax error is encountered.

The parser is a bit sloppy when it comes to reporting the error locations, sometimes pointing farther than he should. I'll see if this is fixable later, or if this is a feature of the lemon parser :smile:

the error message looks like:
```
>> echo '{"statements": ["selecct count(*) from test where x = 12"]}' | http post "127.0.0.1:8080"
{"error":"syntax error around L1:8: `selecct`"}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2362,8 +2362,7 @@ dependencies = [
 [[package]]
 name = "sqlite3-parser"
 version = "0.5.0"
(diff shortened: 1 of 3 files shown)


penberg:
bors merge

PR #96, opened 2023-01-24
Specification for the WebSocket protocol

honzasp:
Specify Hrana, the WebSocket-based protocol for interacting with SQL servers from edge functions.


honzasp:
Maybe it would be better to create a `docs` directory and put the specification and other documentation there? I would like to use this file as a living specification, not as an RFC.

MarinPostma (approved):
bors merge

PR #106, opened 2023-01-25
add http api spec

MarinPostma:
This attempts to specify the HTTP API. This is not a documentation for the current state of the HTTP API, but rather a specification for now, and eventually a documentation.

It brings some strutural changes to the current API, namely:
- top-level errors: since the failure of a statement in a batch should cause the whole batch to rollback, error are move to the top-level. Ultimately this should yield much niecer errors, and a more straighforward behaviour.
- Some changes in the response structure, to leave place for non-backward breaking changes
- Some changes in the encoding of the values, more specifically of integers that Javascript handles badly.

### DIFF
--- a/docs/http_api.md
+++ b/docs/http_api.md
@@ -124,3 +124,11 @@ GET /health
 ```
 
 The health route return an `HTTP 200 (OK)` if the server is up and running.
+
+#### Load dump
+
+```
+POST /load-dump
+```
+
+The load dump allows the user to load a SQLite dump. A dump is created with the `.dump` command in SQLite. This dump can then be sent to sqld with the `POST /load-dump` route, where the body is set to the bytes of the dump file.
--- a/docs/http_api.md
+++ b/docs/http_api.md
@@ -14,9 +14,7 @@ The `Value` type represents an SQLite value. It has 4 variants:
 - Blob: some binary data, encoded in base64
 - Null: the null value.
 
-All these types map to JSON straightforwardly, with two exceptions:
-- Blob: they are represented as an object with { "type": "blob", "base64": /* base64 encoded blob */}
-- Integers: they are represented as string since some languages are not able to represent them in full fidelity.
+All these types map to JSON straightforwardly, except for blobs, that are represented as an object with { "type": "blob", "base64": /* base64 encoded blob */}
 
 ### Response format
 
--- a/docs/http_api.md
+++ b/docs/http_api.md
@@ -46,7 +46,7 @@ Where `T` is the type of the payload in case of success.
 #### Queries
 
 ```
-POST /query
+POST /queries
 ```
 
 This endpoint supports sending batches of queries to the database. All of the statements in the batch are executed as part of a transaction. If any statement in the batch fails, an error is returned and the transaction is aborted, resulting in no change to the database.
--- a/docs/http_api.md
+++ b/docs/http_api.md
@@ -6,17 +6,42 @@ This is the documentation for the sqld HTTP API.
 
 ### The `Value` type
 
-The `Value` type represents a SQLite value. It has 4 variants:
+The `Value` type represents an SQLite value. It has 4 variants:
 
-- Text: an UTF-8 encoded string
-- Integer: a 64-bits signed integer
+- Text: a UTF-8 encoded string
+- Integer: a 64-bit signed integer
 - Real: a 64-bits floating number
 - Blob: some binary data, encoded in base64
 - Null: the null value.
 
 All these types map to JSON straightforwardly, with two exceptions:
 - Blob: they are represented as an object with { "type": "blob", "base64": /* base64 encoded blob */}
-- Integers: they are represented as string, since some languages are not able to represent them in full fidelity.
+- Integers: they are represented as string since some languages are not able to represent them in full fidelity.
+
+### Response format
+
+Responses to queries can either succeed or fail. When they succeed a payload specific to the endpoint being called is returned with a HTTP 200 (OK) status code.
+
+In the case of a failure, a specific `Error` response is returned with the approriate HTTP status code. The `Error` response has the following structure:
+
+```
+type Error = {
+    error: {
+        message: string,
+        error_code: string,
+   }
+}
+```
+
+The error code can later be used to link to the relevant documentation.
+
+The general structure of a response is:
+
+```
+type Response<T> = T | Error;
+```
+
+Where `T` is the type of the payload in case of success.
 
 ### Routes
 
@@ -26,9 +51,9 @@ All these types map to JSON straightforwardly, with two exceptions:
 POST /query
 ```
 
-This endpoint supports sending batches of queries to the database. All of the statements in the batch are executed as part of a transaction. If any statement in the batch fails, an error is returned and the transaction is aborted, resulting in no changed to the database.
+This endpoint supports sending batches of queries to the database. All of the statements in the batch are executed as part of a transaction. If any statement in the batch fails, an error is returned and the transaction is aborted, resulting in no change to the database.
 
-The HTTP API is stateless, this means that interactive transactions are not possible. Since all batches are executed as part of a transactions, any transaction statements (e.g `BEGIN`, `END`, `ROLLBACK`...) are forbidden and will yield an error.
+The HTTP API is stateless, which means that interactive transactions are not possible. Since all batches are executed as part of transactions, any transaction statements (e.g `BEGIN`, `END`, `ROLLBACK`...) are forbidden and will yield an error.
 
 ##### Body
 
@@ -43,14 +68,14 @@ type Query = string | ParamQuery;
 type ParamQuery = { q: string, params: null | Record<string, Value> | Array<Value> }
 ```
 
-Queries are either simple strings, or `ParamQuery` that accept parameter bindings. The `statements` arrays can contain a mix of the two types.
+Queries are either simple strings or `ParamQuery` that accept parameter bindings. The `statements` arrays can contain a mix of the two types.
 
 ##### Response Format
 
-On success, a request to `POST /query` returns a response with a HTTP 200 code, and a JSON body with the following structure:
+On success, a request to `POST /query` returns a response with an HTTP 200 code and a JSON body with the following structure:
 ```
 type BatchResponse = {
-    results: Array<QueryResult> | undefined,
+    results: Array<QueryResult>,
 }
 
 type QueryResult = {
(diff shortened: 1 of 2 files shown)


docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Queries
+
+```
+POST /query
+```
+
+This endpoint supports sending batches of queries to the database. All of the statements in the batch are executed as part of a transaction. If any statement in the batch fails, an error is returned and the transaction is aborted, resulting in no change to the database.
+
+The HTTP API is stateless, which means that interactive transactions are not possible. Since all batches are executed as part of transactions, any transaction statements (e.g `BEGIN`, `END`, `ROLLBACK`...) are forbidden and will yield an error.
```
CodingDoug, line 31: How does this work (or not work) with savepoints? https://www.sqlite.org/lang_savepoint.html

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+
+The body for the query request has the following format:
+
+```
+type QueryBody = {
+    statements: Array<Query>
+}
+
+type Query = string | ParamQuery;
+type ParamQuery = { q: string, params: null | Record<string, Value> | Array<Value> }
```
CodingDoug, line 43: What does null mean here?  Or can we just make the params optional, as they are now in the client TS API?

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+### Error handling
+
+On error, the sqld API returns a unified error payload, with a relevant HTTP status code.
+The error response payload has the following format:
+
+```
+type Error = {
+    error: {
+        message: string,
+        error_code: string,
```
CodingDoug, line 109: Do we have an enumeration of codes yet? Or is this just the same as the HTTP status code?

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+   }
+}
+```
+
+The error code can later be used to link to the relevant documentation.
```
CodingDoug, line 114: Ideally the message can contain a URL with the error code baked into it (fragment) to let the developer go straight there.  We just need to commit to way to host these docs.

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Text: a UTF-8 encoded string
+- Integer: a 64-bit signed integer
+- Real: a 64-bits floating number
+- Blob: some binary data, encoded in base64
+- Null: the null value.
+
+All these types map to JSON straightforwardly, with two exceptions:
+- Blob: they are represented as an object with { "type": "blob", "base64": /* base64 encoded blob */}
+- Integers: they are represented as string since some languages are not able to represent them in full fidelity.
```
CodingDoug, line 19: How does a client discern the type of an integer vs string if they are both encoded as strings?  Maybe everything should be an object like blob with an explicit type.

docs/http_api.md
MarinPostma: Not yet, I will collect them in a separate file once I encounter them in the code

docs/http_api.md
MarinPostma: Yes this is the end goal, but first the codes, then we can add the url field when we have somewhere to take the user to

docs/http_api.md
MarinPostma: I'm not sure about this one tbh, on the one hand typing it is unambiguous, but this is also a significant overhead for integers.  Usually the schema is known, so the user can parse to an integer. If it is not know, parsing to an integer will also succeed, but leaves an ambiguity over the type the field is actually supposed to contain. On the other hand, if it quacks like a duck...

docs/http_api.md
MarinPostma: In the sense that they are optional, yes

CodingDoug: OK, works for me.

docs/http_api.md
CodingDoug: I'm thinking of introspection/visualization tools that we or the developer might build for themselves.

Another thought: if we make the assumption that all values in a column have the same type, maybe the type info should be in the row spec.  Then it won't clutter up the values so much.

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Queries are either simple strings or `ParamQuery` that accept parameter bindings. The `statements` arrays can contain a mix of the two types.
+
+##### Response Format
+
+On success, a request to `POST /query` returns a response with an HTTP 200 code and a JSON body with the following structure:
+```
+type BatchResponse = {
+    results: Array<QueryResult> | undefined,
```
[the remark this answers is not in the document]
CodingDoug: I have been assuming undefined means the field/property simply will not exist if there is no relevant data in it. Unless there is some advantage to always having every possible field and nulling out the ones that don't matter in a given circumstance?

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+    columns: Array<string>,
+    rows: Array<Array<Value>>,
+}
+
+```
+
+On success, `BatchResponse` `error` is undefined, and `results` are populated. This is accompanied by an HTTP 200 code.
+On error, `results` is undefined, and `error` is populated. This is accompanied by the relevant HTTP error code.
+
+Each entry in the `results` array of the `BatchResponse` corresponds to a query in the request.
```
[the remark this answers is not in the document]
CodingDoug: That's what it does now.  The unit tests in place for the client already explicitly check for this to be the case, for both HTTP and Sqlite drivers.

docs/http_api.md
CodingDoug: Textual error messages can be much more descriptive on examination. Take Firebase Auth, for example, with its many many types of errors that would be super difficult to read and keep track of as numbers: https://firebase.google.com/docs/auth/admin/errors

CodingDoug: Right, and I am showing an example where string codes might be easier to read and manage.

And isn't it a SQL tradition to never store a number where no numerical value is being represented? :)

docs/http_api.md
MarinPostma: What would you call it when a field is not defined?

docs/http_api.md
MarinPostma: I also think it should be a string, as mentioned earlier it will be used to generate an url to the relevant documentation

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+
+This is the documentation for the sqld HTTP API.
+
+## Usage
+
+### The `Value` type
+
+The `Value` type represents an SQLite value. It has 4 variants:
+
+- Text: a UTF-8 encoded string
```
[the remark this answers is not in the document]
MarinPostma: Uhmm what would come out of the database when you put a bool in?

docs/http_api.md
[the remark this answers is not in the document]
MarinPostma: Sorry that's an oversight, I used to have the error inside of the batch result, so it was either the error or the results, I will remove it.

By undefined I usually mean that the field can not be present. As opposed to present and null

docs/http_api.md
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type QueryResult = {
+    columns: Array<string>,
+    rows: Array<Array<Value>>,
+}
+
+```
+
+On success, `BatchResponse` `error` is undefined, and `results` are populated. This is accompanied by an HTTP 200 code.
```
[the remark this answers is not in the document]
MarinPostma: Cf earlier comment, forgot the remove it. I thought about putting the error in the batch result, but then opted to make it a generic response on it's own

docs/http_api.md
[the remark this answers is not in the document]
CodingDoug: An integer

```
create table foo (b boolean)
insert into foo values (false)
select * from foo
```
```
B
0
```

This behavior is documented.
https://www.sqlite.org/datatype3.html#boolean_datatype

If something is able to know ahead of time that the column is a boolean, we can either send back a converted boolean or send type information along in the response so the SDK can do it.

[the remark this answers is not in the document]
CodingDoug: We need to figure out the story for date/time types as well.  I haven't played with it enough to know what's best, but it's not immediately intuitive to me.
https://www.sqlite.org/datatype3.html#date_and_time_datatype

docs/http_api.md
MarinPostma: With SQLite being dynamically types, that would probably be confusing... 

@BearLemma stock js json serializer handles that very badly 

With that being said maybe it's the duty of the développer so ensure that he can deserialize safely? We, as the provider of the  SDK, can expose a safe API, would that be a reasonable approach?

CodingDoug: I'd still like to let them know that they're getting something that's not intuitively what they might expect.  If we can tag that row (and perhaps all rows) with type metadata, that would be helpful.

MarinPostma: I'm proposing that we leave integers as is, so it's unambiguous, and handle overflow at the SDK level.

It's not even mandatory to specify a type in the columns in SQLite, so I'm not sure how useful tagging cols would be. We can add it, but it most likely be null, and when not, not to be trusted

docs/http_api.md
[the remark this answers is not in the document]
MarinPostma: that's not the same: the fields don't have the same name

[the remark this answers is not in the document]
MarinPostma: sorry misread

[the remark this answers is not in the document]
MarinPostma: at a hight level, a response in general is:
```
type Response<T> = T  | Error;
```

I'll add a paragraph on that

docs/http_api.md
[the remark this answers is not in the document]
MarinPostma: @BearLemma we don't have the same guarantees as chiselstrike, the columns need not be typed. At this point I think it's up to the user's ORM to figure this out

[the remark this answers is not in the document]
MarinPostma: we can provide column affinity _if we have it_, columns could be of type:
```
type Column = { name: string, type: string | undefined };
```

docs/http_api.md
MarinPostma: I think they should be supported as is, but I'll confirm when I have dig further into the issue

CodingDoug: It might be kind of odd for the developer to work with savepoints in a way that appears to be outside of a transaction (from their POV, since they are never allowed to be explicit about it).  I am leaning toward disallowing anything that has to do with a transaction.

The downside to this is that now they can't do anything with transactions other than implied transaction we provide.

MarinPostma: also, it's not like they can really perform any logic, so maybe it's not very usefull

PR #112, opened 2023-01-26
Sketch the Kysely adapter for sqld

honzasp:
Using the Hrana client from `@libsql/hrana-client`, this PR sketches an adapter ("dialect") for Kysely.


packages/kysely-sqld/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "@libsql/kysely-sqld",
+
+  "type": "module",
+  "main": "lib-cjs/index.js",
+  "types": "lib-esm/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./lib-esm/index.d.ts",
+      "import": "./lib-esm/index.js",
+      "require": "./lib-cjs/index.js"
```
MarinPostma, line 11: I don't get why sometimes things come from the `lib-esm` dir, sometimes, from the `lib-cjs`?

MarinPostma (approved):
looks good

packages/kysely-sqld/package.json
honzasp: ES modules come from `lib-esm`, and CJS modules come from `lib-cjs`. TypeScript declarations are the same regardless of the target module system, so they are produced only in `lib-esm`.

honzasp: They are emitted here by the two invocations of `tsc` from `scripts.build-{cjs,esm}`, using `tsconfig.build-{cjs,esm}.json` as tsconfig.

honzasp:
@penberg You are right, thanks for pointing out, fixed!

MarinPostma:
bors merge

PR #113, opened 2023-01-26
open replication hook db out of the main thread

MarinPostma:
`ReplicationHook` was opening it's database on the main thread. That was all fine until libsql started to call `xFrames` on open, which caused the call to `block_on` in `xFrame` to panic


MarinPostma:
bors merge

Issue #65, opened 2023-01-19
It should be an error to try to run more than one statement with client lib `execute()`


CodingDoug:
If a statement string is issued with more than one command separated by semicolons, everything but the first command is silently ignored.  Instead, I think sqld should reject the query with an error saying that each string can only contain one statement.

CodingDoug:
@MarinPostma I edited the description here.  No idea how it said something other than what I changed it to (maybe copy/pasta?).

The turso shell exhibits this problem as well if I issue multiple statements at once with a semicolon. I presume it is just passing along the given command string directly to the HTTP interface.

MarinPostma:
@CodingDoug we can do that. It is SQLite behavior to only take the first statement into account, and ignore anything that follows, but Postgres returns an error saying it got more results than It was expecting.

CodingDoug:
@BearLemma I believe that sqld is using a sqlite parser called lemon-rs.
https://github.com/gwenn/lemon-rs

So, I would assume that after using that to parse, it would be possible to check how many statements were found.

MarinPostma:
@BearLemma you can change the behaviour here: https://github.com/libsql/sqld/blob/main/sqld/src/http/mod.rs#L112-L115

PR #119, opened 2023-01-27
Allow only one command per statement in HTTP interface..

BearLemma:
Fixes https://github.com/libsql/sqld/issues/65

@MarinPostma @CodingDoug I've fixed it for HTTP. The user now gets an error response 
```
{
	"error": "found more than one command in a single statement string. It is allowed to issue only one command per string."
}
```

But we use that parser from multiple places. Are multiple commands per statement acceptable in those scenarios? 
Other usages:
https://github.com/libsql/sqld/blob/main/sqld/src/rpc/proxy.rs#L153
https://github.com/libsql/sqld/blob/main/sqld/src/postgres/proto.rs#L65
https://github.com/libsql/sqld/blob/main/sqld/src/postgres/proto.rs#L100

### DIFF
--- a/sqld/src/http/mod.rs
+++ b/sqld/src/http/mod.rs
@@ -109,10 +109,13 @@ fn error(msg: &str, code: StatusCode) -> Response<Body> {
 fn parse_queries(queries: Vec<QueryObject>) -> anyhow::Result<Vec<Query>> {
     let mut out = Vec::with_capacity(queries.len());
     for query in queries {
-        let stmt = Statement::parse(&query.q)
-            .next()
-            .transpose()?
-            .unwrap_or_default();
+        let mut iter = Statement::parse(&query.q);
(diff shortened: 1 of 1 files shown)


MarinPostma:
the rest is fine, is posgres protocol, let's focus on HTTP for now

sqld/src/query_analysis.rs
```diff
@@ -96,7 +96,7 @@ impl Statement {
                 kind,
             })
         }
-        let mut parser = Box::new(Parser::new(s.as_bytes()));
+        let mut parser = Parser::new(s.as_bytes());
```
MarinPostma, line 99: why did you do that?
The parser needs to be boxed because it's very big. please undo that, and I'll add a comment

MarinPostma (approved):
bors merge

PR #134, opened 2023-01-31
Improve Error handling and error propagation.

BearLemma:
Rename QueryError to SqldError.
Use SqldError wherever possible to better propagate errors.

### DIFF
--- a/sqld/src/database/libsql.rs
+++ b/sqld/src/database/libsql.rs
@@ -9,12 +9,11 @@ use rusqlite::OpenFlags;
 use tokio::sync::oneshot;
 use tracing::warn;
 
+use crate::error::Error;
 use crate::libsql::wal_hook::WalHook;
-use crate::query::{
-    Column, ErrorCode, Params, Queries, Query, QueryError, QueryResponse, QueryResult, ResultSet,
-    Row,
-};
+use crate::query::{Column, Params, Queries, Query, QueryResponse, QueryResult, ResultSet, Row};
 use crate::query_analysis::{State, Statement};
+use crate::Result;
 
 use super::{Database, TXN_TIMEOUT_SECS};
 
@@ -48,7 +47,7 @@ fn execute_query(conn: &rusqlite::Connection, stmt: &Statement, params: Params) 
 
     params
         .bind(&mut prepared)
-        .map_err(|e| QueryError::new(ErrorCode::SQLError, e))?;
+        .map_err(Error::LibSqlInvalidQueryParams)?;
 
     let mut qresult = prepared.raw_query();
 
@@ -202,7 +201,7 @@ impl LibSqlDb {
             Arc<Mutex<mwal::ffi::libsql_wal_methods>>,
         >,
         wal_hook: impl WalHook + Send + Clone + 'static,
-    ) -> anyhow::Result<Self> {
+    ) -> crate::Result<Self> {
         let (sender, receiver) = crossbeam::channel::unbounded::<Message>();
 
         tokio::task::spawn_blocking(move || {
@@ -245,12 +244,7 @@ impl LibSqlDb {
                 } else {
                     // fail all the queries in the batch with timeout error
                     let errors = (0..queries.len())
-                        .map(|_| {
-                            Err(QueryError::new(
-                                ErrorCode::TxTimeout,
-                                "transaction timedout",
-                            ))
-                        })
+                        .map(|idx| Err(Error::LibSqlTxTimeout(idx)))
                         .collect();
                     ok_or_exit!(resp.send((errors, state.state)));
                     timedout = false;
@@ -264,7 +258,7 @@ impl LibSqlDb {
 
 #[async_trait::async_trait]
 impl Database for LibSqlDb {
-    async fn execute(&self, queries: Queries) -> anyhow::Result<(Vec<QueryResult>, State)> {
+    async fn execute(&self, queries: Queries) -> Result<(Vec<QueryResult>, State)> {
         let (resp, receiver) = oneshot::channel();
         let msg = Message { queries, resp };
         let _ = self.sender.send(msg);
--- a/sqld/src/http/mod.rs
+++ b/sqld/src/http/mod.rs
@@ -5,6 +5,7 @@ use std::future::poll_fn;
 use std::net::SocketAddr;
 use std::sync::Arc;
 
+use anyhow::Context;
 use base64::prelude::BASE64_STANDARD_NO_PAD;
 use base64::Engine;
 use bytes::{BufMut, Bytes, BytesMut};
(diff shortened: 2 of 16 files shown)


sqld/src/database/libsql.rs
```diff
@@ -264,11 +258,11 @@ impl LibSqlDb {
...
 #[async_trait::async_trait]
 impl Database for LibSqlDb {
-    async fn execute(&self, queries: Queries) -> anyhow::Result<(Vec<QueryResult>, State)> {
+    async fn execute(&self, queries: Queries) -> Result<(Vec<QueryResult>, State)> {
         let (resp, receiver) = oneshot::channel();
         let msg = Message { queries, resp };
         let _ = self.sender.send(msg);
 
-        Ok(receiver.await?)
+        receiver.await.map_err(Error::LibSqlReceiverError)
```
MarinPostma, line 266: I think Receive error can be collapsed to an internal error, type erased error. You can implement `From<tokio::RecvError> for Error` so you don't have to`map_err` and simply use `?`

sqld/src/http/mod.rs
```diff
@@ -250,7 +255,7 @@ where
         }
     });
 
-    server.await?;
+    server.await.map_err(Error::HttpServerError)?;
```
MarinPostma, line 258: this is one instance where anyhow is fine. This error will never be exposed to the client, but will cause the app to exit. We'll never need to give it an error code.

sqld/src/lib.rs
```diff
@@ -94,7 +98,7 @@ where
     }
 
     while let Some(res) = handles.next().await {
-        res??;
+        res.map_err(|e| Error::Internal(e.to_string()))??;
```
MarinPostma, line 101: Same here, anyhow is fine here.

sqld/src/lib.rs
```diff
@@ -149,7 +153,8 @@ pub async fn run_server(config: Config) -> Result<()> {
                 #[cfg(feature = "mwal_backend")]
                 vwal_methods,
             )
-            .await?;
+            .await
+            .map_err(Error::ProxyError)?;
```
MarinPostma, line 157: anyhow is fine here as well: we are initiating the proxy server, and failing here means that the app will exit with an error

sqld/src/lib.rs
```diff
@@ -38,6 +39,8 @@ pub enum Backend {
     Mwal,
 }
 
+type Result<T> = std::result::Result<T, Error>;
```
MarinPostma, line 42: this should probably live in the `error` module.

sqld/src/database/libsql.rs
```diff
@@ -48,7 +47,7 @@ fn execute_query(conn: &rusqlite::Connection, stmt: &Statement, params: Params)
 
     params
         .bind(&mut prepared)
-        .map_err(|e| QueryError::new(ErrorCode::SQLError, e))?;
+        .map_err(Error::LibSqlInvalidQueryParams)?;
```
MarinPostma, line 50: This error could be returned from the bind function directly

sqld/src/lib.rs
```diff
@@ -82,7 +85,8 @@ where
     handles.push(tokio::spawn(server.serve(factory)));
 
     if let Some(addr) = config.http_addr {
-        let authorizer = http::auth::parse_auth(config.http_auth)?;
+        let authorizer =
+            http::auth::parse_auth(config.http_auth).map_err(Error::AuthConfigError)?;
```
MarinPostma, line 89: the `parse_auth` method should return the appropriate error

sqld/src/rpc/mod.rs
```diff
@@ -43,13 +45,16 @@ where
         let tls_config = tonic::transport::ServerTlsConfig::new()
             .identity(identity)
             .client_ca_root(ca_cert);
-        builder = builder.tls_config(tls_config)?;
+        builder = builder
+            .tls_config(tls_config)
+            .map_err(|e| Error::TlsConfigError(e.to_string()))?;
```
MarinPostma, line 50: This error should cause the app to exit, anyhow is ok

sqld/src/rpc/mod.rs
```diff
@@ -43,13 +45,16 @@ where
...
     }
     builder
         .add_service(ProxyServer::new(proxy_service))
         .add_service(WalLogServer::new(logger_service))
         .serve(addr)
-        .await?;
+        .await
+        .map_err(|e| Error::Internal(format!("RPC server exited with an error: {e}")))?;
```
MarinPostma, line 57: same here: anyhow is fine

sqld/src/rpc/proxy.rs
```diff
@@ -30,43 +31,26 @@ pub mod proxy_rpc {
...
+        fn from(other: SqldError) -> Self {
             match other {
-                ErrorCode::SqlError => crate::query::ErrorCode::SQLError,
-                ErrorCode::TxBusy => crate::query::ErrorCode::TxBusy,
-                ErrorCode::TxTimeout => crate::query::ErrorCode::TxTimeout,
-                ErrorCode::Internal => crate::query::ErrorCode::Internal,
+                SqldError::LibSqlInvalidQueryParams(_) => ErrorCode::SqlError,
+                SqldError::LibSqlTxTimeout(_) => ErrorCode::TxTimeout,
+                SqldError::LibSqlTxBusy => ErrorCode::TxBusy,
+                _ => ErrorCode::Internal,
```
MarinPostma, line 49: we're losing type information over the proxy :/  We need to think about that.

sqld/src/rpc/proxy.rs
```diff
@@ -103,12 +87,18 @@ pub mod proxy_rpc {
         }
     }
 
+    impl From<bincode::Error> for SqldError {
+        fn from(other: bincode::Error) -> Self {
+            Self::Internal(other.to_string().into())
```
MarinPostma, line 92:
```suggestion
            Self::Internal(other.to_string())
```

MarinPostma, line 92: maybe we should implement all conversion methods inside in the `error` module?

sqld/src/server/mod.rs
```diff
@@ -47,7 +47,7 @@ impl Server {
         Ok(self)
     }
 
-    pub async fn serve<S>(self, mut make_svc: S) -> anyhow::Result<()>
+    pub async fn serve<S>(self, mut make_svc: S) -> Result<()>
```
MarinPostma, line 50: anyhow error is good here

sqld/src/wal_logger.rs
```diff
@@ -136,7 +137,7 @@ impl WalLogger {
     /// Size of the file header
     pub const HEADER_SIZE: usize = 4096;
 
-    pub fn open(path: impl AsRef<Path>) -> anyhow::Result<Self> {
+    pub fn open(path: impl AsRef<Path>) -> Result<Self> {
```
MarinPostma, line 140: I think anyhow is fine here as well

MarinPostma (approved):
bors merge

PR #151, opened 2023-02-02
Include content-type header for query response.

BearLemma:
One thing I'm not sure about is whether it's necessary to include the encoding in the header as well :thinking:


MarinPostma (approved):
bors merge

PR #152, opened 2023-02-02
fix HTTP blob encoding

MarinPostma:
This PR fixes and unify blob encoding in http payloads as

```json
{ "base64": "PAYLOAD" }
```

@penberg I'm wondering if we shouldn't parse the blob into an ArrayBuffer?


MarinPostma:
bors merge

Issue #145, opened 2023-02-01
Make sqlite dependency optional in JavaScript client


penberg:
As suggested by @CodingDoug, let's move the sqlite dependency in `optionalDependencies` so that you can use the client with just HTTP.

CodingDoug:
See also #121

PR #154, opened 2023-02-06
Use sql.js to provide sqlite driver in browser.

BearLemma:
It doesn't use webworkers as I can't imagine anyone using it in the browser for anything but testing. 
Fixes https://github.com/libsql/sqld/issues/145.

### DIFF
--- a/packages/libsql-client/src/lib/driver/BrowserSqliteDriver.ts
+++ b/packages/libsql-client/src/lib/driver/BrowserSqliteDriver.ts
@@ -1,0 +1,82 @@
+import initSqlJs from "sql.js";
+import { BoundStatement, Params, ResultSet, SqlValue } from "../libsql-js";
+import { Driver } from "./Driver";
+
+export class BrowserSqliteDriver implements Driver {
+    private sql?: initSqlJs.SqlJsStatic;
+    private db?: initSqlJs.Database;
+
+    constructor(url: string) {
+        if (url !== "file::memory:" && url !== ":memory:") {
+            console.warn(
+                `BrowserSqliteDriver will ignore given db url '${url}' as in browser mode only the memory storage is available`
+            );
+        }
+    }
+
+    private async loadWasm() {
+        this.sql = await initSqlJs({
+            // Required to load the wasm binary asynchronously. Of course, you can host it wherever you want
+            // You can omit locateFile completely when running in node
+            locateFile: (file) => "https://cdnjs.cloudflare.com/ajax/libs/sql.js/1.8.0/sql-wasm.js"
+        });
+        this.db = new this.sql.Database();
+    }
+
+    async execute(sql: string, params?: Params): Promise<ResultSet> {
+        if (this.sql === undefined) {
+            await this.loadWasm();
+        }
+
+        return await new Promise((resolve) => {
+            let columns: string[];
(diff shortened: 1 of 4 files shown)


packages/libsql-client/src/lib/driver/BrowserSqliteDriver.ts
```diff
@@ -0,0 +1,74 @@
+import  initSqlJs from "sql.js";
+import { BoundStatement, Params, ResultSet, SqlValue } from "../libsql-js";
+import { Driver } from "./Driver";
+
+export class BrowserSqliteDriver implements Driver {
+    private sql?: initSqlJs.SqlJsStatic;
+    private db?: initSqlJs.Database;
+
+    constructor(url: string) {
+        if (url !== "file::memory:") {
```
CodingDoug, line 10: This can be ":memory:" as well.

https://www.sqlite.org/inmemorydb.html

packages/libsql-client/src/lib/driver/BrowserSqliteDriver.ts
```diff
@@ -0,0 +1,74 @@
...
+            locateFile: file => 'https://cdnjs.cloudflare.com/ajax/libs/sql.js/1.8.0/sql-wasm.js'
+        });
+        this.db = new this.sql.Database();
+    }
+
+    async execute(sql: string, params?: Params): Promise<ResultSet> {
+        if (this.sql === undefined) {
+            await this.loadWasm();
+        }
+        console.log(sql);
```
CodingDoug, line 28: This looks like debug logging

packages/libsql-client/src/lib/libsql-js.ts
```diff
@@ -77,6 +78,12 @@ export function connect(config: Config): Connection {
     if (url.protocol == "http:" || url.protocol == "https:") {
         return new Connection(new HttpDriver(url));
     } else {
-        return new Connection(new SqliteDriver(rawUrl));
+        // Hack taken from https://stackoverflow.com/a/31090240/1474847
+        const isBrowser = new Function("try {return this===window;}catch(e){ return false;}");
```
CodingDoug, line 82: There are some other hacks that will go in the other direction, creating and populating a `window` in order to  polyfill web APIs on node. This would seem incompatible with those.

Maybe the browser-or-node package recommended as a different answer would be better?

packages/libsql-client/src/lib/driver/BrowserSqliteDriver.ts
```diff
@@ -0,0 +1,74 @@
...
+        const result = [];
+        for (const stmt of stmts) {
+            let rs;
+            if (typeof stmt === "string") {
+                rs = await this.execute(stmt);
+            } else {
+                rs = await this.execute(stmt.q, stmt.params);
+            }
+            result.push(rs);
+        }
```
CodingDoug, line 71: Wrap in begin/commit?

packages/libsql-client/src/lib/libsql-js.ts
CodingDoug: How about: https://stackoverflow.com/a/31456668/807126

CodingDoug (changes requested):

CodingDoug (approved):

PR #155, opened 2023-02-06
Make SQLite node driver take params with prefix.

BearLemma:
It's a bit of a hack, but it should work reasonably well and we can still enjoy the benefits of better-sqlite speed.


CodingDoug (approved):

CodingDoug:
bors merge

PR #159, opened 2023-02-07
Handshake

haaawk:


sqld/src/rpc/mod.rs
```diff
@@ -21,14 +23,15 @@ pub async fn run_rpc_server<F>(
     ca_cert_path: Option<PathBuf>,
     factory: F,
     logger: Arc<WalLogger>,
+    hello_response: HelloResponse,
```
MarinPostma, line 26: I think that this information should be retrieved from the `WalLogger`

haaawk: done.

sqld/src/lib.rs
```diff
@@ -167,7 +172,30 @@ pub async fn run_server(config: Config) -> anyhow::Result<()> {
...
+                Err(e) if e.kind() == ErrorKind::UnexpectedEof => {
+                    // OK: the file was empty, this is first generation
+                    0
+                }
+                Err(e) => Err(e)?,
+            };
+            generation_file
+                .write_all_at(&generation_id.to_le_bytes(), 0)
+                .unwrap();
+
```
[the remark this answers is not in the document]
MarinPostma: I think `WalLogger` should have a `generation` method that returns  the generation and the start frame index

sqld/src/rpc/wal_log.rs
```diff
@@ -3,18 +3,25 @@ pub mod wal_log_rpc {
...
 
-use self::wal_log_rpc::{wal_log_entry::Payload, Frame, LogOffset, WalLogEntry as RpcWalLogEntry};
+use self::wal_log_rpc::{
+    wal_log_entry::Payload, Frame, HelloRequest, HelloResponse, LogOffset,
+    WalLogEntry as RpcWalLogEntry,
+};
 
 pub struct WalLogService {
     logger: Arc<WalLogger>,
+    replicas_with_hello: RwLock<Vec<SocketAddr>>,
```
MarinPostma, line 23: what about an `HashSet` here?

sqld/src/rpc/wal_log.rs
```diff
@@ -74,8 +85,32 @@ impl WalLog for WalLogService {
...
+        req: tonic::Request<HelloRequest>,
+    ) -> Result<tonic::Response<HelloResponse>, Status> {
+        let replica_addr = req
+            .remote_addr()
+            .ok_or(Status::internal("No remote RPC address"))?;
+        {
+            let mut guard = self.replicas_with_hello.write().await;
+            if !guard.contains(&replica_addr) {
+                guard.push(replica_addr);
+            }
```
MarinPostma, line 112: let's just use an hashset to make this operation idempotent

sqld/src/rpc/wal_log.rs
```diff
@@ -74,8 +85,32 @@ impl WalLog for WalLogService {
         &self,
         req: tonic::Request<LogOffset>,
     ) -> Result<tonic::Response<Self::LogEntriesStream>, Status> {
+        let replica_addr = req
+            .remote_addr()
+            .ok_or(Status::internal("No remote RPC address"))?;
+        {
+            let guard = self.replicas_with_hello.read().await;
+            if !guard.contains(&replica_addr) {
+                return Err(Status::failed_precondition("Hello message missing"));
```
MarinPostma, line 94: can we use something that look more like an error code, like `"NO_HELLO"`, and put that in a `const` so we can reuse that to check for the code on the client side

MarinPostma (changes requested):

sqld/src/rpc/wal_log.rs
haaawk: HashSet will perform worse unless we have thousands of replicas. Both cpu wise and memory wise.

sqld/src/rpc/wal_log.rs
haaawk: hashset is an overkill when we expect << 100 replicas

MarinPostma: it's just more ergonomic, and expresses the intent better

sqld/src/rpc/wal_log.rs
MarinPostma: "early optimization is the root of all evil" 😄 This is clearly not on the hot path

sqld/src/rpc/wal_log.rs
```diff
@@ -42,7 +52,15 @@ impl From<(u64, WalLogEntry)> for RpcWalLogEntry {
 
 impl WalLogService {
     pub fn new(logger: Arc<WalLogger>) -> Self {
-        Self { logger }
+        let first_frame_id_in_generation: u64 = logger.get_current_offset().try_into().unwrap();
+        Self {
+            logger,
+            replicas_with_hello: RwLock::new(HashSet::<SocketAddr>::new()),
+            hello_response: HelloResponse {
+                generation_id: uuid::Uuid::new_v4().to_string(),
```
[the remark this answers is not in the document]
MarinPostma: that would make it more consistent to how this is serialized in other protocols: https://github.com/libsql/sqld/blob/main/sqld/proto/proxy.proto#L6-L7

sqld/src/rpc/wal_log.rs
```diff
@@ -74,8 +92,36 @@ impl WalLog for WalLogService {
...
         req: tonic::Request<LogOffset>,
     ) -> Result<tonic::Response<Self::LogEntriesStream>, Status> {
+        let replica_addr = req
+            .remote_addr()
+            .ok_or(Status::internal("No remote RPC address"))?;
+        {
+            let guard = self
+                .replicas_with_hello
+                .read()
+                .map_err(|_| Status::internal("Lock error"))?;
```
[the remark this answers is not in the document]
haaawk: fixed.

sqld/src/rpc/wal_log.rs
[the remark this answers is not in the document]
haaawk: done

MarinPostma (approved):
bors merge

PR #163, opened 2023-02-07
genId check function

MarinPostma:
small function to check if a (genId, start_frame_id) is compatible with a replica

### DIFF
--- a/sqld/src/database/write_proxy/replication.rs
+++ b/sqld/src/database/write_proxy/replication.rs
@@ -80,6 +80,7 @@ impl PeriodicDbUpdater {
 
 struct ReadReplicationHook {
     logger: WalLogClient<Channel>,
+    current_gen: u64,
     fetch_frame_index: u64,
     /// Persistent last committed index used for restarts.
     /// The File should contain two little-endian u64:
@@ -252,6 +253,8 @@ impl ReadReplicationHook {
             last_applied_index,
             buffer: Default::default(),
             rt: Handle::current(),
+            // TODO: read from file
+            current_gen: 0,
         })
     }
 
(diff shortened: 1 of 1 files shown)


sqld/src/database/write_proxy/replication.rs
```diff
@@ -252,6 +253,8 @@ impl ReadReplicationHook {
             last_applied_index,
             buffer: Default::default(),
             rt: Handle::current(),
+            // TODO: read from file
```
haaawk, line 256: I think for the first version we can even live without persisting this on the replica.

sqld/src/database/write_proxy/replication.rs
```diff
@@ -80,6 +80,7 @@ impl PeriodicDbUpdater {
 
 struct ReadReplicationHook {
     logger: WalLogClient<Channel>,
+    current_gen: u64,
```
haaawk, line 83: shouldn't this be uuid (string or u128)?

sqld/src/database/write_proxy/replication.rs
```diff
@@ -305,4 +308,29 @@ impl ReadReplicationHook {
 
         Ok(())
     }
+
+    #[allow(dead_code)]
+    fn is_valid_new_gen(&mut self, gen: u64, gen_start_frame_index: u64) -> bool {
```
haaawk, line 313: shouldn't `gen` be uuid (string or u128)?

sqld/src/database/write_proxy/replication.rs
```diff
@@ -305,4 +308,29 @@ impl ReadReplicationHook {
...
+        if self.current_gen != gen {
+            if let Some(last_applied) = self.last_applied_index {
+                if last_applied <= gen_start_frame_index {
+                    // TODO persist to file.
```
haaawk, line 317: I think for the first iteration it's totally fine to just keep current_gen in memory. Replicas don't use persistent disks so if the machine is restarted we probably loose this info anyway.

sqld/src/database/write_proxy/replication.rs
```diff
@@ -305,4 +308,29 @@ impl ReadReplicationHook {
...
+                    self.current_gen = gen;
```
haaawk, line 318: I think it would be cleaner to update current_gen outside of this function. Same way restart should also be triggered outside based on the result of this function.

haaawk (commented):
LGTM in general. Left some comments.

sqld/src/database/write_proxy/replication.rs
MarinPostma: That's true for turso, but sqld should target more general deployment scenario, given that it's an open source project

haaawk: Good point. Still probably not absolutely required in the first iteration.

MarinPostma:
closing, mostly rewritten in #177

PR #177, opened 2023-02-09
Replication: full handshake protocol

MarinPostma:
implement the replication protocol handshake

### DIFF
--- a/sqld/proto/wal_log.proto
+++ b/sqld/proto/wal_log.proto
@@ -2,14 +2,19 @@ syntax = "proto3";
 package wal_log;
 
 message LogOffset {
-    uint64 startOffset = 1;
+    // offset of the last reveiced frame. If absent, the server should send frame 0.
+    optional uint64 currentOffset = 1;
 }
 
 message HelloRequest { }
 
 message HelloResponse {
+    /// Uuid of the current generation
     string generation_id = 1;
-    uint64 first_frame_id_in_generation = 2;
+    /// First frame index in the current generation
+    uint64 generation_start_index = 2;
+    /// Uuid of the database being replicated
+    string database_id = 3;
 }
 
 message Frame {
--- a/sqld/src/database/write_proxy/replication.rs
+++ b/sqld/src/database/write_proxy/replication.rs
@@ -21,24 +21,33 @@ use std::io::ErrorKind;
 use std::mem::size_of;
 use std::os::unix::prelude::FileExt;
 use std::path::Path;
+use std::str::FromStr;
 use std::time::Duration;
 
-use anyhow::ensure;
+use anyhow::Context;
+use bytemuck::{bytes_of, try_from_bytes, Pod, Zeroable};
+use crossbeam::channel::TryRecvError;
 use futures::StreamExt;
 use rusqlite::ffi::SQLITE_ERROR;
 use rusqlite::OpenFlags;
 use tokio::runtime::Handle;
 use tonic::transport::Channel;
+use tonic::Code;
+use uuid::Uuid;
 
 use crate::libsql::ffi::{types::XWalFrameFn, PgHdr, Wal};
 use crate::libsql::open_with_regular_wal;
 use crate::libsql::wal_hook::WalHook;
 use crate::rpc::wal_log::wal_log_rpc::{wal_log_client::WalLogClient, LogOffset};
+use crate::rpc::wal_log::wal_log_rpc::{HelloRequest, HelloResponse};
+use crate::rpc::wal_log::NO_HELLO_ERROR_MSG;
 use crate::wal_logger::{WalFrame, WAL_PAGE_SIZE};
+use crate::HARD_RESET;
 
 pub struct PeriodicDbUpdater {
     interval: Duration,
     db: rusqlite::Connection,
+    abort_receiver: crossbeam::channel::Receiver<ReplicationError>,
 }
 
 /// The `PeriodicUpdater` role is to periodically trigger a dummy write that will be intercepted by
@@ -49,7 +58,7 @@ impl PeriodicDbUpdater {
         logger: WalLogClient<Channel>,
         interval: Duration,
     ) -> anyhow::Result<Self> {
-        let hook = ReadReplicationHook::new(logger, path).await?;
+        let (hook, abort_receiver) = ReadReplicationHook::new(logger, path).await?;
         let path = path.to_owned();
         let db = tokio::task::spawn_blocking(move || {
             open_with_regular_wal(
@@ -64,23 +73,42 @@ impl PeriodicDbUpdater {
         })
         .await??;
 
-        Ok(Self { interval, db })
+        Ok(Self {
+            interval,
+            db,
+            abort_receiver,
+        })
     }
 
     /// blocking!
-    pub fn step(&mut self) {
-        // dummy write that triggers a call to xFrame
-        let _ = self.db.execute(
-            "create table if not exists __dummy__ (dummy); insert into __dummy__ values (1);",
-            (),
-        );
-        std::thread::sleep(self.interval);
+    /// return whether to continue
+    pub fn step(&mut self) -> anyhow::Result<bool> {
+        match self.abort_receiver.try_recv() {
+            Ok(e) => {
+                // received an error from loop
+                Err(e.into())
+            }
+            Err(TryRecvError::Empty) => {
+                // dummy write that triggers a call to xFrame
+                let _ = self.db.execute(
+                    "create table if not exists __dummy__ (dummy); insert into __dummy__ values (1);",
+                    (),
+                );
+                std::thread::sleep(self.interval);
+
+                Ok(true)
+            }
+
+            Err(TryRecvError::Disconnected) => {
+                // graceful exit
+                Ok(false)
+            }
+        }
     }
 }
 
 struct ReadReplicationHook {
     logger: WalLogClient<Channel>,
-    fetch_frame_index: u64,
     /// Persistent last committed index used for restarts.
     /// The File should contain two little-endian u64:
     /// - The first one is the attempted commit index before the call xFrame
@@ -88,11 +116,16 @@ struct ReadReplicationHook {
     /// After a flight of pages has been successfully written, the two numbers should be the same.
     /// On startup the two number are checked for consistency. If they differ, the database is
     /// considered corrupted, since it is impossible to know what the actually replicated index is.
-    last_applied_index_file: File,
-    last_applied_index: Option<u64>,
+    wal_index_meta_file: File,
+    wal_index_meta: WalIndexMeta,
     /// Buffer for incoming frames
     buffer: VecDeque<WalFrame>,
     rt: Handle,
+    /// A channel to send error back to the polling loop.
+    /// When an error occurs that causes an abort, this handle should be replaced with None, and
+    /// the error sent. This means that if `abort_sender` is None, we should assume a previous
+    /// abort.
+    abort_sender: Option<crossbeam::channel::Sender<ReplicationError>>,
 }
 
 /// Debug assertion. Make sure that all the pages have been applied
@@ -122,6 +155,11 @@ unsafe impl WalHook for ReadReplicationHook {
         sync_flags: c_int,
         orig: XWalFrameFn,
     ) -> c_int {
+        // A fatal error has occured before, don't do anyhting
+        if self.abort_sender.is_none() {
+            return SQLITE_ERROR;
+        }
+
         let rt = self.rt.clone();
         if let Err(e) = rt.block_on(self.fetch_log_entries()) {
             tracing::error!("error fetching log entries: {e}");
@@ -133,24 +171,14 @@ unsafe impl WalHook for ReadReplicationHook {
             assert_ne!(size_after, 0, "commit index points to non commit frame");
 
             tracing::trace!(commit = ?frame_count, size_after = size_after);
-            let attempted_commit_index = self
-                .last_applied_index
-                .map(|x| x + frame_count as u64)
-                .unwrap_or_default();
-            // pre-write index
-            self.last_applied_index_file
-                .write_all_at(&attempted_commit_index.to_le_bytes(), 0)
-                .unwrap();
-
+            self.inc_pre_commit(frame_count as u64)
+                .expect("failed to write pre-commit index");
             let ret = unsafe { orig(wal, WAL_PAGE_SIZE, page_headers, size_after, 1, sync_flags) };
 
             if ret == 0 {
                 debug_assert!(all_applied(page_headers));
-                // persist new commited index
-                self.last_applied_index_file
-                    .write_all_at(&attempted_commit_index.to_le_bytes(), size_of::<u64>() as _)
-                    .unwrap();
-                self.last_applied_index.replace(attempted_commit_index);
+                self.sync_post_commit()
+                    .expect("failed to write post-commit index");
                 // remove commited entries.
                 self.buffer.drain(..frame_count);
                 tracing::trace!("applied frame batch");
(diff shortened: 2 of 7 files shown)


haaawk:
Is all this stuff needed on a single PR? I'm especially pointing to:
```
add sqld default-run
set default log level to info
fix MWAL feature
improve WAL log file structure and protocol
```
Bundling too much unrelated changes in a single PR drastically decreases the quality of the review.

sqld/src/database/write_proxy/replication.rs
```diff
@@ -80,6 +80,7 @@ impl PeriodicDbUpdater {
 
 struct ReadReplicationHook {
     logger: WalLogClient<Channel>,
+    current_gen: u64,
```
haaawk: Same comment as https://github.com/libsql/sqld/pull/163#discussion_r1101303323

sqld/src/database/write_proxy/replication.rs
```diff
@@ -305,4 +308,29 @@ impl ReadReplicationHook {
 
         Ok(())
     }
+
+    #[allow(dead_code)]
+    fn is_valid_new_gen(&mut self, gen: u64, gen_start_frame_index: u64) -> bool {
```
haaawk: Same comment as https://github.com/libsql/sqld/pull/163#discussion_r1101303681

sqld/src/database/write_proxy/replication.rs
```diff
@@ -305,4 +308,29 @@ impl ReadReplicationHook {
...
+        if self.current_gen != gen {
+            if let Some(last_applied) = self.last_applied_index {
+                if last_applied <= gen_start_frame_index {
+                    // TODO persist to file.
+                    self.current_gen = gen;
```
haaawk: Same comment as https://github.com/libsql/sqld/pull/163#discussion_r1101306608

sqld/src/database/write_proxy/replication.rs
MarinPostma: those do not exist anymore

sqld/src/rpc/wal_log.rs
```diff
@@ -24,7 +24,7 @@ pub struct WalLogService {
     replicas_with_hello: RwLock<HashSet<SocketAddr>>,
 }
 
-const _NO_HELLO_ERROR_MSG: &str = "NO_HELLO";
```
haaawk, line 27: This commit should probably come later in the history to keep it bisectable.

sqld/proto/wal_log.proto
```diff
@@ -2,21 +2,26 @@ syntax = "proto3";
 package wal_log;
 
 message LogOffset {
-    uint64 startOffset = 1;
+    // offset of the last reveiced frame. If absent, the server should send frame 0.
+    optional uint64 currentOffset = 1;
```
haaawk, line 6: This is not backwards compatible, is it? If a primary with a previous version receives an empty LogOffset, it will not know how to handle it, will it?

MarinPostma: no it's not backward compatible

sqld/proto/wal_log.proto
```diff
@@ -2,21 +2,26 @@ syntax = "proto3";
...
 
 message Frame {
     bytes data = 1;
 }
 
 service WalLog {
-    rpc Hello(HelloRequest) returns (HelloResponse) {}
-    rpc LogEntries(LogOffset) returns (stream Frame) {}
+    rpc Hello(HelloRequest) returns (HelloResponse) { }
+    rpc LogEntries(LogOffset) returns (stream Frame) { }
```
haaawk, line 26: Changes like that are such a useless distractions :(.

sqld/src/rpc/wal_log.rs
```diff
@@ -72,7 +72,8 @@ impl WalLog for WalLogService {
                 return Err(Status::failed_precondition(NO_HELLO_ERROR_MSG));
             }
         }
-        let start_offset = req.into_inner().start_offset;
+        // if current_offset is None, then start sending from 0, otherwise return next frame
+        let start_offset = req.into_inner().current_offset.map(|x| x + 1).unwrap_or(0);
```
haaawk, line 76: So this is a change in the behaviour. Was it a bug that we were sending frame from `start_offset` not `start_offset+1` before?

sqld/src/rpc/wal_log.rs
```diff
@@ -88,6 +89,12 @@ impl WalLog for WalLogService {
             let mut guard = self.replicas_with_hello.write().unwrap();
             guard.insert(replica_addr);
         }
-        Ok(tonic::Response::new(self.logger.hello_response().clone()))
+        let response = HelloResponse {
+            database_id: self.logger.database_id.to_string(),
+            generation_start_index: self.logger.generation.start_index,
+            generation_id: self.logger.generation.id.to_string(),
```
haaawk, line 95: Maybe we could store generation.id as string and avoid calling to_string every time? I know it's not performance critical but it's not that storing a string would be worse in any way.

sqld/src/rpc/wal_log.rs
```diff
@@ -88,6 +89,12 @@ impl WalLog for WalLogService {
             let mut guard = self.replicas_with_hello.write().unwrap();
             guard.insert(replica_addr);
         }
-        Ok(tonic::Response::new(self.logger.hello_response().clone()))
+        let response = HelloResponse {
+            database_id: self.logger.database_id.to_string(),
```
haaawk, line 93: Unless someone else is using database_id, maybe we could just store it as string? I know it's not performance critical but there's no cost in storing it as string so it's better to have more performant code.

sqld/proto/wal_log.proto
haaawk: Can we make it backward compatible? We should avoid breaking changes like this when not absolutely necessary.

sqld/src/rpc/wal_log.rs
MarinPostma: no, before the client was asking for the frame that was after the last one it had received, now it tell the server what was the last frame it received and the server sends the next

sqld/src/rpc/wal_log.rs
MarinPostma: you'd have to clone it, so that does not make any difference

sqld/src/database/write_proxy/replication.rs
```diff
@@ -207,43 +200,149 @@ fn free_page_header(h: *const PgHdr) {
...
+    /// attempts to merge two meta files.
+    fn merge_from_hello(mut self, hello: HelloResponse) -> Result<Self, MetaMergeError> {
+        let hello_db_id = Uuid::from_str(&hello.database_id)
+            .context("invalid database id from primary")?
+            .as_u128();
+        let hello_gen_id = Uuid::from_str(&hello.generation_id)
+            .context("invalid generation id from primary")?
+            .as_u128();
+
+        if hello_db_id == self.database_id {
```
haaawk: Shouldn't it be != instead of ==?

sqld/src/database/write_proxy/replication.rs
```diff
@@ -207,43 +200,149 @@ fn free_page_header(h: *const PgHdr) {
...
+            return Err(MetaMergeError::DbIncompatible);
+        }
+
+        if self.generation_id == hello_gen_id {
+            return Ok(self);
+        } else if self.generation_id != hello_gen_id
```
haaawk, line 296: Isn't `self.generation_id != hello_gen_id` guaranteed in this else branch?

sqld/src/rpc/wal_log.rs
haaawk: It seems that converting number to string takes more work than just copying an array of bytes which works great with caches.

sqld/src/database/write_proxy/replication.rs
MarinPostma: oops yeah I changed this to explicitly test test this error being triggered and forgot to change it back

sqld/src/rpc/wal_log.rs
haaawk: I see. This is really breaking new/old version compatibility.

sqld/src/database/write_proxy/replication.rs
```diff
@@ -207,43 +200,149 @@ fn free_page_header(h: *const PgHdr) {
...
+            None => Ok(WalIndexMeta::new_from_hello(hello)?),
+        }
+    }
+
+    fn flush_meta(&self) -> anyhow::Result<()> {
+        self.wal_index_meta_file
+            .write_all_at(bytes_of(&self.wal_index_meta), 0)?;
+
+        Ok(())
+    }
```
haaawk, line 392: I'm just wondering, shouldn't this be async? It seems to be disk IO.

sqld/src/database/write_proxy/replication.rs
```diff
@@ -251,12 +253,12 @@ impl WalIndexMeta {
             .context("invalid generation id from primary")?
             .as_u128();
 
-        if hello_db_id == self.database_id {
+        if hello_db_id != self.database_id {
```
haaawk, line 290: It seems that you fixed it here @MarinPostma

sqld/src/database/write_proxy/replication.rs
MarinPostma: oh wait actually it's solved later

sqld/src/database/write_proxy/replication.rs
MarinPostma: yes, I also think it makes the condition clearer if someone ever moves the branches around

sqld/src/database/write_proxy/replication.rs
MarinPostma: it is called from a sync context

sqld/src/database/write_proxy/replication.rs
MarinPostma: yes, I was using this to artificially trigger the error while testing

sqld/src/rpc/wal_log.rs
```diff
@@ -90,11 +90,11 @@ impl WalLog for WalLogService {
             guard.insert(replica_addr);
         }
         let response = HelloResponse {
-            database_id: self.logger.database_id.to_string(),
+            database_id: dbg!(self.logger.database_id.to_string()),
```
haaawk: We probably want this `dbg!(` removed :)

sqld/src/wal_logger.rs
```diff
@@ -246,18 +246,26 @@ impl WalLogger {
         let file_end = log_file.metadata()?.len();
         let end_id;
         let current_checksum;
+
         let header = if file_end == 0 {
+            let db_id = Uuid::new_v4();
+            dbg!(db_id);
+            dbg!(Uuid::from_u128(db_id.as_u128()));
```
haaawk: Debugging leftovers?

sqld/src/wal_logger.rs
```diff
@@ -271,6 +279,7 @@ impl WalLogger {
             let header = WalLoggerFileHeader::decode(&header_buf)?;
             end_id = (file_end - size_of::<WalFrameHeader>() as u64) / Self::FRAME_SIZE as u64;
             current_checksum = AtomicU64::new(Self::compute_checksum(&header, &log_file)?);
+            dbg!(header.db_id);
```
haaawk: Another one.

sqld/src/wal_logger.rs
```diff
@@ -246,18 +246,26 @@ impl WalLogger {
...
-                db_id: Uuid::new_v4().as_u128(),
+                db_id: db_id.as_u128(),
             };
 
             let mut header_buf = BytesMut::new();
             header.encode(&mut header_buf);
+            let decoded = WalLoggerFileHeader::decode(&header_buf).unwrap();
+
+            dbg!(decoded.db_id);
+            dbg!(header.db_id);
```
haaawk: and here

sqld/src/rpc/wal_log.rs
MarinPostma: they're gone later

sqld/src/database/write_proxy/replication.rs
```diff
@@ -375,22 +378,34 @@ impl ReadReplicationHook {
...
 
-        Ok(())
+                Ok(())
+            }
+            Err(s) if s.code() == Code::FailedPrecondition && s.message() == NO_HELLO_ERROR_MSG => {
+                tracing::info!("Primary restarted, perfoming hanshake again");
+                self.wal_index_meta =
+                    Self::perform_handshake(Some(self.wal_index_meta), self.logger.clone()).await?;
+
+                Ok(())
```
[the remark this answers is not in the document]
MarinPostma: yes, next iteration will

sqld/src/database/write_proxy/replication.rs
haaawk: I don't think it's useful to duplicate checks like this. If someone moves branches around they have to make sure the logic is correct. Moving branches around without thinking is never a good idea :). Let's remove this redundant check.

sqld/src/database/write_proxy/replication.rs
haaawk: I see.

haaawk:
Looks mostly good but I don't think we have to break backwards compatibility here.

sqld/src/rpc/wal_log.rs
MarinPostma: maybe from a memcpy standpoint, but this is very likely trumped by the cost of allocation anyways. I'm deeply convinced that the performance gain from converting to a string here is insignificant compared to the gains of preserving type information

sqld/proto/wal_log.proto
MarinPostma: the protocol is not backward compatible, since it now requires a handshake

sqld/src/rpc/wal_log.rs
haaawk: The value of preserving type information is 0. No one uses this in form other than string.

It does not matter that some other cost trumps this. In well written systems there are no clear bottlenecks and everything takes 0.5-2% of performance. Then it becomes clear that small things add up to the slowness.

If there was a downside of storing this as string as oppose to UUID, it would be reasonable to ignore the performance but since there's absolutely no value in keeping this as UUID - I don't think it's worth keeping the value in suboptimal way.

sqld/proto/wal_log.proto
haaawk: Yes but this change makes it impossible to do rolling upgrade of the cluster.
Normally you would update replicas one by one and then update the primary - no downtime for reads, short downtime for writes.

It's a good practise to not require the whole cluster to be turned down for upgrade.

sqld/src/rpc/wal_log.rs
MarinPostma: [A Uuid *is* a u128](https://en.wikipedia.org/wiki/Universally_unique_identifier), a string is only a representation of it. So when storing as a string we are storing a representation of a Uuid, not a u128. The representation can vary, but the Uuid it is representing is always the same. There are many reasons to use the Uuid type, over its string representation:
- it's Copy, whereas a string is not
- it lives on the stack, whereas a string is on the heap
- It's always comparable
- It's self-documenting
- It has a known size
- it comes with a full library to work with
- it is easily serializable in a zero-copy format
- it's typed

I have more arguments if this does not convince you :)

sqld/proto/wal_log.proto
MarinPostma: You have a plan for rolling upgrades with only the handshake?

haaawk: Yes. Were you planning to shut everything down for update?

sqld/src/rpc/wal_log.rs
haaawk: > it's Copy, whereas a string is not

Do we ever copy/clone it? It seems that we put it into the meta object and it stays there for ever while being accessed through references, no?

>it lives on the stack, whereas a string is on the heap

Are you sure? https://docs.rs/uuid/latest/src/uuid/lib.rs.html#436

>It's always comparable

Are we ever comparing it on the primary?

> It's self-documenting

This is a fair point but using a `generation_uuid` name would be even more self documenting

> It has a known size

Are you sure? https://docs.rs/uuid/latest/src/uuid/lib.rs.html#436

> it comes with a full library to work with

Are we using any functions from that library?

>it is easily serializable in a zero-copy format

When we're serializing it to string it's both an allocation and a copy of the content, no?

>it's typed

What benefit does that give when we always pass it as string in the end?

sqld/proto/wal_log.proto
haaawk: So we leave databases for existing ~100 users broken?

The way I envision it:
1. We don't break proto compatibility
2. New version of the replica should gracefully handle if hello message is rejected and assume everything is good in such case
3. We upgrade all replicas and then switch the primary

I think that would work, no?

sqld/src/rpc/wal_log.rs
MarinPostma: > Are you sure? https://docs.rs/uuid/latest/src/uuid/lib.rs.html#436
absolutely: https://docs.rs/uuid/latest/src/uuid/lib.rs.html#283

Let's look at string representation for a bit. There are at least 4 common and accepted ways to represent a Uuid
- braced: `{6aacd448-3de1-4e37-832c-6e48b8bb2ca0} `: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.as_braced
- hyphenated: `6aacd448-3de1-4e37-832c-6e48b8bb2ca0`: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.as_hyphenated
- simple: `6aacd4483de14e37832c6e48b8bb2ca0`: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.as_simple
- urn: `urn:uuid:203e7fa8-65d6-46a6-ba25-72e7b222ed99 `: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.as_urn
Now what that means is that if you want to compare them, you always need to parse them into a valid `Uuid` instance.

Let's say that a developer comes to the codebase. They are unaware of this optimization, and you see that the id is a string, how confident are you that they will go with the correct (and not obvious) way, that is convert back to Uuid, and compare them, or make sure that the representation is consistent with the one that was previously there, and that someone else relying on? What about other platforms? what about a uuid you receive from some other place? How do other platforms decide to encode it? Are you confident that such a mistake is caught in review?

We can also talk about perf, I have run some benches:

```
alloc string            time:   [20.232 ns 20.250 ns 20.272 ns]
                        change: [-0.5384% -0.3590% -0.1514%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 16 outliers among 100 measurements (16.00%)
  2 (2.00%) high mild
  14 (14.00%) high severe

Benchmarking to_string: Collecting 100 samples in estimated 5.0001 s (146M ite
                                                                              
to_string               time:   [33.970 ns 34.010 ns 34.069 ns]
                        change: [+0.6582% +0.8872% +1.1115%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

At first glance, It is true that converting to string is slower (34ns vs 20ns), but let's look a the fuller picture here, and consider what the actual cost/benefits are, for let's say 10000 replicas:
- assuming that the handshakes are serialized (of course they run concurrently), 10000*15ns = 150us in the lifetime of the cluster. After a full day we're looking at a (1.5e-8 / 86400) * 100 = 1.7361111e-11% improvement
- On the other hand, we get code that is a potential foot gun, and if the developer is careful enough, he will have to jump back and forth between the Uuid and its representation

haaawk: I don't agree with the developer argument. All the following questions are irrelevant in our situation:
```
What about other platforms? what about a uuid you receive from some other place? How do other platforms decide to encode it?
```
The field is set using Uuid so it's very easy to see what the format is. I don't expect anyone to try to parse such string without first looking at a place where the string is created :).

The second argument is valid though. Even though it's partially incorrect - it's 150us per primary restart not in the lifetime of the cluster. It's still very low so ok - let's keep it the way it is. I still don't see the reason to keep it as UUID and redo the work over and over again but since you proved the benefit is small - let it be.

sqld/proto/wal_log.proto
MarinPostma: let's consider the two scenarios:
1. we first update the primary and then replicas:
If we do that, the primary is going to return `NO_HELLO` when old replicas try to replicate. We could make the primary ignore when the replica hasn't made a handshake, but how do you differentiate an old replica that does not know about hello, and a new one that hasn't performed a handshake? This makes the handshake pointless
2. we upgrade replicas first and then primary:
replicas will send a handshake to the server and the server will return an error. What you suggest is to proceed with asking for frames anyway, but that kills the purpose of the handshake again. The purpose of the handshake is for the replica and the primary to agree on the data that will be exchanged.

haaawk: So I was thinking about (2). I don't think it defeats the purpose of the handshake. If the primary returns error on the handshake then we assume it's the old primary that does not understand the hello message. As long as we distinguish errors returned from the new primary from "don't understand the message" error, it all should be good and the replica should be able to decide if it deals with an old primary or new primary with error.

MarinPostma: In the current state, the primary cannot be upgraded. The database directory structure has been changed in #94, the wal format has been changed in #72 on which this PR is based, along with the format of the frame being sent, and other changes to the protocol. We are working on protocol improvements because we can't even assume that restarting the primary is safe at this point. 

Moreover, I thought we agreed that breaking was fine at his point during the previous town hall, and it is under this assumption that #94 was merged.

haaawk: My understanding is that we can break things when there's no other way and when it gives us in return something valuable.

We should start thinking about doing changes in such a way that would allow us to upgrade existing clusters. We have ~100 users now and loosing even few of them because of not being able to fix their clusters is a very very bad thing.

I'm particularly unhappy with the fact that we have existing clusters that have broken replication and now we won't be able to fix them.

@penberg my understanding was that we start doing changes in a more responsible way with at least a migration path. Has anything changed?

haaawk: So https://github.com/libsql/sqld/pull/94 shouldn't prevent us from upgrading the primary - can't we just move the db file to the new dir and start the primary in a new version?

haaawk: https://github.com/libsql/sqld/pull/72 is not yet merged so maybe we should first fix the replication and only then merge https://github.com/libsql/sqld/pull/72. Is this PR using anything but database_id from https://github.com/libsql/sqld/pull/72 @MarinPostma ?

haaawk: It seems that database_id check is not crucial for replication so we can:
1. fix replication
2. do https://github.com/libsql/sqld/pull/72
3. Add database_id check to handshake

no?

I think it is important to fix existing clusters and not just ignore them if we can and it seems that we can.

MarinPostma: we can, we just need to do it - in the same way we could manually dump all existing cluster and re-import them in the new version

haaawk: it sucks but maybe that's what we should do. That brings us back to the backwards compatibility issue

haaawk: Resolving this as we agreed on the dump - load migration procedure.

MarinPostma:
bors merge

Issue #179, opened 2023-02-09
cross-fetch in the JavaScript driver doesn't seem to work everywhere


penberg:
We've had multiple people point out that they had to switch to native `fetch()` because of some cross-fetch issue.

penberg:
Looks like this `cross-fetch` just ignores native fetch: https://github.com/lquixada/cross-fetch/issues/69

CodingDoug:
I think we can keep using cross-fetch, but don't import it as `fetch` (which masks the native fetch). Then, check to see if there is a native fetch to use before falling back to cross-fetch.

PR #181, opened 2023-02-10
JS client: use native fetch when available.

BearLemma:
Might be fixing https://github.com/libsql/sqld/issues/179 
Needs testing though.


packages/libsql-client/src/lib/driver/HttpDriver.ts
```diff
@@ -25,11 +25,17 @@ export class HttpDriver implements Driver {
...
             body: JSON.stringify(statements)
-        });
+        };
+
+        let response;
+        if (typeof fetch === "function") {
+            response = await fetch(this.url, reqParams);
+        } else {
+            response = await crossFetch(this.url, reqParams);
+        }
```
CodingDoug, line 38: It might be more elegant to express it like this instead?

```
const f = typeof fetch === "function" ? fetch : crossFetch
const response = await f(...)
```

CodingDoug:
> Yeah I tried that but the compiler gave me a finger :(

Really?  tsc gave me no issues with that.

MarinPostma:
@BearLemma this may overlap with #214 where I had to switch to axios because of the keep-alive mess.

CodingDoug:
> @BearLemma this may overlap with #214 where I had to switch to axios because of the keep-alive mess.

@MarinPostma Does axios work on CloudFlare?  If it's not using native fetch behind the scenes, it might not work on CF at all. CF workers are one of our primary use case right now for this client library.

MarinPostma:
uhh no it does not. So we abandon support for node?
Let's make sure that worker fetch supports `keep-alive` at least

MarinPostma:
what a joke this language is...

CodingDoug:
> uhh no it does not. So we abandon support for node? Let's make sure that worker fetch supports `keep-alive` at least

Not sure what you mean. This library already works on node, browsers, and react native via the cross-fetch library.  This PR enables the use of native fetch if it's available (which cross-fetch inconveniently does not delegate to correctly).

MarinPostma:
I'm referring to #214. I need to find a way to make the version check portable.

PR #183, opened 2023-02-10
sqlite dump

MarinPostma:
This PR allows sqlite dumps to be loaded over HTTP.
For now it requires that the dump fits in memory.

you can create a dump to the `sqlite-dump` file, from a sqlite db with the following commands:
```
> .out sqlite-dump
> .dump
```

and you can load it to an existing sqlite instance with:
```
POST /load-dump
```


MarinPostma:
bors merge

PR #198, opened 2023-02-13
Raw statements and results in Hrana client

honzasp:
This is useful for data mapper libraries which introduce their own mapping between JavaScript and SQLite types.

The disadvantage is that it exposes protocol-level details in the public API, but the protocol should be at least as stable as the library itself, so this should not be a big problem. If it is, we can mark the `executeRaw()` method as unstable.


MarinPostma (approved):
bors merge

PR #205, opened 2023-02-14
scale to zero

MarinPostma:
This PR adds a `--idle-shutdown-timeout-s <secs>` arg to sqld that shuts down the server after <secs> of inactivity on the http server.

### DIFF
--- a/sqld/src/http/services/idle_shutdown.rs
+++ b/sqld/src/http/services/idle_shutdown.rs
@@ -1,0 +1,74 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use tokio::sync::{watch, Notify};
+use tokio::time::timeout;
+use tower::{Layer, Service};
+
+pub struct IdleShutdownLayer {
+    watcher: Arc<watch::Sender<()>>,
+}
+
+impl IdleShutdownLayer {
+    pub fn new(idle_timeout: Duration, shutdown_notifier: Arc<Notify>) -> Self {
+        let (sender, mut receiver) = watch::channel(());
+        tokio::spawn(async move {
+            loop {
+                // FIXME: if we measure that this is causing performance issues, we may want to
+                // implement some debouncing.
+                let timeout_fut = timeout(idle_timeout, receiver.changed());
+                match timeout_fut.await {
+                    Ok(Ok(_)) => continue,
+                    Ok(Err(_)) => break,
+                    Err(_) => {
+                        shutdown_notifier.notify_waiters();
+                    }
+                }
+            }
+        });
+
+        Self {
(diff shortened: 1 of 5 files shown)


sqld/src/lib.rs
```diff
@@ -85,6 +86,7 @@ pub struct Config {
     pub rpc_server_ca_cert: Option<PathBuf>,
     pub enable_bottomless_replication: bool,
     pub create_local_http_tunnel: bool,
+    pub idle_shutdown_timeout: Option<Duration>,
```
haaawk, line 91: I've seen in the past `_s` or `_seconds` being added to time params name to make is more apparent what unit is used. Do you think it would make sense to add it to our name as well?

haaawk: Nvm. I see you did that on the command-line param.

sqld/src/http/services/idle_shutdown.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub fn new(idle_timeout: Duration, shutdown_notifier: Arc<Notify>) -> Self {
+        let (sender, mut receiver) = watch::channel(());
+        tokio::spawn(async move {
+            loop {
+                // FIXME: if we measure that this is causing performance issues, we may want to
+                // implement some debouncing.
+                let timeout_fut = timeout(idle_timeout, receiver.changed());
+                match timeout_fut.await {
+                    Ok(Ok(_)) => continue,
+                    Ok(Err(_)) => break,
```
haaawk, line 22: Is it guaranteed that the server will shutdown when hit this branch? I'm trying to assess if it would be possible to hit this branch and the instance not shutting down. That would mean a working instance that never scales down, no? Unless I'm misunderstanding something.

Maybe it makes sense to call `shutdown_notifier.notify_waiters()` here as well?

haaawk: or not break out of the loop at least

haaawk:
This is very cool. Thanks for doing it so quickly! Why is it a draft? Is there any remaining work left?

sqld/src/http/services/idle_shutdown.rs
MarinPostma: This branch is hit when the sender has be dropped on the other side, so we won't be receiving messages anymore. I think it's safe to assume that the server is already shutting down in this case. (It could also be a hard reset)

haaawk: 👍 thanks for the explanation. Would it make sense to to at least log a warning in case we ever have to debug somewhere around here?

MarinPostma (commented):
bors merge

PR #209, opened 2023-02-15
fix(sqld): env var for backend config

chrisgacsal:
Fix environment variable name for `sqld` backend configuration.


MarinPostma (approved):
Good catch! Thanks for your contribution :)

bors merge

PR #213, opened 2023-02-15
replace custom logging layer with http-tower TraceLayer

MarinPostma:
This pr replace the custom `LoggerLayer` with `tower-http` provided `TraceLayer`


MarinPostma (commented):
bors merge

PR #214, opened 2023-02-15
js sdk version check

MarinPostma:
Check server version to select query URL.

### DIFF
--- a/packages/js/libsql-client/src/http/http-driver.ts
+++ b/packages/js/libsql-client/src/http/http-driver.ts
@@ -3,9 +3,15 @@ import { ResultSet, BoundStatement, Params } from "../shared-types.js";
 import { Driver } from "./../driver.js";
 import { Base64 } from "js-base64";
 
+const compatibleFetch = typeof fetch === "function" ? fetch : crossFetch;
+const versionMatcher = /(\d+).(\d+).(\d+)/;
+
+type Version = [number, number, number];
+
 export class HttpDriver implements Driver {
     private url: URL;
     private authHeader: string | undefined;
+    private version: Version | undefined;
 
     constructor(url: URL) {
         if (url.username !== "" || url.password !== "") {
@@ -44,8 +50,8 @@ export class HttpDriver implements Driver {
             };
         }
 
-        const compatibleFetch = typeof fetch === "function" ? fetch : crossFetch;
-        const response = await compatibleFetch(this.url, reqParams);
+        const queryUrl = await this.queryUrl();
+        const response = await compatibleFetch(queryUrl, reqParams);
         if (response.status === 200) {
             const results = await response.json();
             validateTopLevelResults(results, statements.statements.length);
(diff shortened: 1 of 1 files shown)


MarinPostma:
@athoscouto @BearLemma Could you review this? I hope my JS is not too broken.

MarinPostma:
@haaawk this will break the client API, since connect needs to be `async`

MarinPostma:
@haaawk NVM I found a wait to make it non-breaking.

packages/js/libsql-client/src/http/http-driver.ts
```diff
@@ -69,6 +75,42 @@ export class HttpDriver implements Driver {
             throw new Error(`${response.status} ${response.statusText}`);
         }
     }
+
+    private async queryUrl(): Promise<URL> {
+        if (this.version === undefined) {
+            this.version = await this.getVersion();
+            return this.queryUrl();
+        }
```
CodingDoug, line 83: Caching the version has one downside: since the client is stateless, the version could actually change on the server between requests. In this case, the user of this library has an inaccurate view of the actual version that could be in play for the next request.

packages/js/libsql-client/src/http/http-driver.ts
```diff
@@ -3,9 +3,15 @@ import { ResultSet, BoundStatement, Params } from "../shared-types.js";
 import { Driver } from "./../driver.js";
 import { Base64 } from "js-base64";
 
+const compatibleFetch = typeof fetch === "function" ? fetch : crossFetch;
+const versionMatcher = /(\d+).(\d+).(\d+)/;
+
+type Version = [number, number, number];
```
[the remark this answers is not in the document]
MarinPostma: you're right this is much better, I'll update that

packages/js/libsql-client/src/http/http-driver.ts
MarinPostma: I think that is overkill for now. This is only useful when a server upgrades to a breaking version. But in that case, there isn't much to do but update the SDK as well. It would improve the DX for sure, bat as far as the end user is concerned, it's just gonna break, and he'll be as clueless.

The only case where it's important to know that the version changed is on a server version downgrade, where we should be switching compatibility mode, but I don't think that it's worth the effort for now, wdyt?

CodingDoug: To be honest, upon closer look, I'm not clear what the intended use cases are for this automatic version polling. Right now it seems to:

1. Intercept the first request and make the version available by API _after_ everything finishes.
2. Or fail the entire request if there was no correctly formatted version.
3. Or fail the entire request if there was something other than a 200 or 404.

For 1, no value seems to be added to the original request that triggered the version polling - just the overhead of making an extra request.  I have to follow up with a second API call to `getVersion` get the data. And that followup call requires yet another poll to the server to get the version (it doesn't return the cached version)

For 2 and 3, if I'm seeing one of these failures, what does that mean to me, as the developer? What do I do to overcome them? Or am I being protected from something catastrophic in all this?

packages/js/libsql-client/src/http/http-driver.ts
```diff
@@ -69,6 +75,42 @@ export class HttpDriver implements Driver {
...
+            return [versionNums[0], versionNums[1], versionNums[2]];
+        }
+
+        // handle pre-version route
+        if (resp.status == 404) {
+            // dummy version < to any other version.
+            return [0, 0, 0]
+        }
+
+        throw new Error(await resp.text());
```
CodingDoug, line 112: What is this error going to tell me, the caller? Will it be specific or actionable?  It's going to preempt the original request that triggered it, which might be confusing in that context.

packages/js/libsql-client/src/http/http-driver.ts
MarinPostma: the only reason this exists is for the newer clients to be compatible with the old version of the server, this is apparently a prerequisite for migrating from `/` to `/queries`. What we want to cover with this is a beta user that would have upgraded its client but not its server, and that would break.

So yeah, I agree, that creates latency, it's possibly confusing, but we want non-breakage at all costs...

MarinPostma: Or we could accept just breaking beta software, and people will understand because it's beta software

CodingDoug: I personally have zero concerns about anything with a major version number of 0 breaking in order to accommodate expansion.  And I'd also rather have the server version number fed to me in every response. :)

MarinPostma: But that does not solve our issue with picking the route for querying based on the version, so that's probably out of the scope of this pr

MarinPostma: I'm fine with adding a custom header as well of course

penberg:
The client moved to another repository: https://github.com/libsql/libsql-client-ts

I am closing this pull request. Please open a new one if it's still relevant.

PR #216, opened 2023-02-15
move js SDKs to separate directory

MarinPostma:
housekeeping


MarinPostma:
bors merge

honzasp:
Maybe we should also move `pypackages` to `packages/py`, then?

PR #220, opened 2023-02-15
remove jepsen boilerplate

MarinPostma:
housekeeping.

This is just boilerplate, and no actual tests exist. Let's bring them back if we think we need then later on.


MarinPostma:
bors merge

PR #225, opened 2023-02-16
add release drafter to github workflow

MarinPostma:
automatically create release drafts


MarinPostma:
it's ok, release drafter is expecting some file in main that I'm adding in this pr.

bors merge

PR #226, opened 2023-02-16
Small simplifications

honzasp:
- Replace some generics with dynamic polymorphism
- Replace `FuturesUnordered<JoinHandle>` with a `JoinSet`
- Drop `FutOrNever`: this reimplements `Fuse` from the `futures` crate, and does not seem to be used anyway
- Drop some unnecessary `Sync` constraints (it is almost never legitimate to require a future to be `Sync`)

These are just small local simplifications that I found when starting the work on Hrana server that I think can be merged separately to avoid polluting the main PR.


honzasp:
bors merge (after an informal approval from @MarinPostma)

PR #230, opened 2023-02-17
python package check version

MarinPostma:
Add version check to python package

### DIFF
--- a/packages/python/libsql-client/libsql_client/http_driver.py
+++ b/packages/python/libsql-client/libsql_client/http_driver.py
@@ -1,26 +1,59 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple, TypeAlias
+from urllib.parse import urljoin
 import aiohttp
 import base64
 import json
 
 from .driver import _Driver, _RawStmt
-from .errors import ClientResponseError, ClientHttpError
+from .errors import ClientResponseError, ClientHttpError, ClientError
 from .result import ResultSet, Row, Value
+
+Version: TypeAlias = Tuple[int, int, int]
 
 class _HttpDriver(_Driver):
     _session: aiohttp.ClientSession
     _url: str
+    _version: Optional[Version]
 
     def __init__(self, url: str) -> None:
         self._session = aiohttp.ClientSession()
         self._url = url
+        self._version = None
+
+    async def _get_version(self) -> Version:
+        url = urljoin(self._url, "version")
+        async with await self._session.get(url) as resp:
+            if not resp.ok:
+                if resp.status == 404:
+                    # pre /version, return dummy 0.0.0 version
+                    return 0, 0, 0
+                resp_body = await resp.read()
+                try:
+                    message = json.loads(resp_body).get("error")
+                except ValueError:
+                    message = None
+                raise ClientHttpError(resp.status, message)
+            version_string = await resp.text()
+            parts = [int(x) for x in version_string.split(".")]
+            if len(parts) != 3:
+                raise ClientError("server returned invalid version number")
+            return tuple(parts)
+
+    async def _get_or_load_cached_version(self) -> Version:
+        if self._version is None:
+            self._version = await self._get_version()
+        return self._version
+
+
 
     async def batch(self, stmts: List[_RawStmt]) -> List[ResultSet]:
         req_body = {
             "statements": [_encode_stmt(stmt) for stmt in stmts],
         }
 
-        async with await self._session.post(self._url, json=req_body) as resp:
+        version = await self._get_or_load_cached_version()
+        url = self._queries_url(version)
+        async with await self._session.post(url, json=req_body) as resp:
             if not resp.ok:
                 resp_body = await resp.read()
                 try:
@@ -39,6 +72,10 @@ class _HttpDriver(_Driver):
 
     async def close(self) -> None:
         await self._session.close()
+
+    # TODO: use version to dispatch query route
+    def _queries_url(self, _version: Version):
+        return urljoin(self._url, "")
 
 def _encode_stmt(stmt: _RawStmt) -> Any:
     params_json: Any


packages/python/libsql-client/tests/test_client.py
```diff
@@ -37,14 +37,6 @@ async def test_response_error(url):
             await client.execute("SELECT foo")
         assert "no such column: foo" in str(excinfo.value)
 
-@pytest.mark.asyncio
-async def test_http_invalid_url_error(http_url):
-    invalid_url = http_url + "/invalid/url"
-    async with libsql_client.Client(invalid_url) as client:
-        with pytest.raises(libsql_client.ClientHttpError) as excinfo:
-            await client.execute("SELECT 1")
-        assert "404" in str(excinfo.value)
```
honzasp, line 46: I assume that this test needed to be removed because we are now ignoring the path in the URL passed to the `Client`. Maybe we should not do this, and use relative paths in the `Client`?

packages/python/libsql-client/libsql_client/http_driver.py
```diff
@@ -1,26 +1,56 @@
...
     _session: aiohttp.ClientSession
     _url: str
+    _version: Optional[tuple[int, int, int]]
 
     def __init__(self, url: str) -> None:
         self._session = aiohttp.ClientSession()
         self._url = url
+        self._version = None
+
+    async def session(self):
```
honzasp, line 21: In Python, the convention is to prefix private methods with `_`. You can resolve the clash with `_session` by renaming it to `_get_session()` or, even better, to `_get_version()` (and make it return the version, not the session).

packages/python/libsql-client/libsql_client/http_driver.py
```diff
@@ -1,26 +1,56 @@
...
 
 from .driver import _Driver, _RawStmt
-from .errors import ClientResponseError, ClientHttpError
+from .errors import ClientResponseError, ClientHttpError, ClientError
 from .result import ResultSet, Row, Value
 
 class _HttpDriver(_Driver):
     _session: aiohttp.ClientSession
     _url: str
+    _version: Optional[tuple[int, int, int]]
```
honzasp, line 14: As a convention, the package uses types from `typing` (i.e. `Tuple`), instead of the builtin types (i.e. `tuple`), to retain compatibility with Python versions before 3.9. (Not that this compatibility is tested, though...)

packages/python/libsql-client/libsql_client/http_driver.py
```diff
@@ -40,6 +70,9 @@ async def batch(self, stmts: List[_RawStmt]) -> List[ResultSet]:
     async def close(self) -> None:
         await self._session.close()
 
+    def queries_url(self):
```
honzasp, line 73: Ditto, it would be better to prefix the private method with `_`. Also, perhaps we should add the logic for switching between `/` and `/queries` here?

honzasp (changes requested):

packages/python/libsql-client/libsql_client/http_driver.py
MarinPostma: yes this is a placeholder for switching the route, but the queries route is not yet merged, and this will be merged first

packages/python/libsql-client/libsql_client/http_driver.py
MarinPostma: thanks

packages/python/libsql-client/libsql_client/http_driver.py
honzasp: In that case, perhaps we should merge this PR later? At this point, it's just a noop that introduces extra latency.

packages/python/libsql-client/libsql_client/http_driver.py
MarinPostma: I already have a `_get_version` method, the goal here is to lazy retrieve the version number when an HTTP call is made.

MarinPostma: so one should be using the `get_sessions()` method instead of the `_session` field

honzasp: I understand your intent, but there are two issues:
- Private methods should be prefixed with an underscore
- The functionality of this method has nothing to do with the session, it deals with the server version, so it should be named accordingly and it should return an appropriate value

MarinPostma: private method addressed

what do you suggest? the goal is to ensure that the session is only accessed with an adequately initialized driver. would moving the logic that sets the version to methods like `_ensure_client_init` make sense to you?

honzasp: I would suggest renaming the method to `_get_cached_version()` or something similar and let it return the version as `Tuple[int, int, int]`. You don't need to call this method every time you need to use the session (e.g. in `_get_version()`!), you need to call this method when you need to _know the server version_. The type system would nudge you: `self._version` is `Optional[Tuple]`, while the result from `_ensure_version()` would be just a `Tuple`.

(The "cached" in the name would also serve as a hint that the version is cached and may be invalid. This is the reason why HTTP APIs typically don't rely on clients querying the server version, but they put the burden of compatibility on the servers: the client specifies the version it expects.

A typical solution is that HTTP servers expose versioned paths, and the clients target the specific version that they expect. For an example, see `iku-turso-api`, which exposes `/v1` and `/v2` versions.)

MarinPostma: yeah, we're not going down the version API road just now, or we'll be on v40 by the end of the month.
The goal here is to allow a new client to work with an old DB
I'm not calling `_get_version()` every time, but only once per session
I don't care where the check is done, but it has to be done once, preferably somewhere where the control flow converges

MarinPostma: This check can be removed when users are migrated. 

We could also crash and return that the version of the server is incompatible with the client, and that it should be updated

MarinPostma: You have a clear idea of what you want, so, what do we do now to let old db work with new client? Just tell me, I'll do it

honzasp: If we want to keep compatiblity with old server, why don't we just keep using the "/" path in the client? This is what the PR would do anyway: we read the version from the server, but we don't use it.

In other words, if we care about compatibility, we should first upgrade the servers to support the new API in addition to the old API, then upgrade the clients to use the new API unconditionally.

If we don't care about compatibility, just switch both the server and the client to the new API.

MarinPostma: When we know what will be the version that contains the `/queries` route, we can check the version in the client. For now, we're not certain what will be the version number when we release `/queries`.

The proposed PR for `/queries` handled redirection of `/` to `/queries`, but what is not possible is to make old servers redirect `/queries` to `/`, and this is what we're trying to achieve here. The part you are describing is already covered.

MarinPostma: I just want to know how you suggest we handle the case where an old server (supports *only* `/`) is used with a new client (queries `/queries`)

honzasp: I suggest that we introduce new clients (using `/queries`) only after we upgrade our servers to handle both `/` and `/queries`.

MarinPostma: Server are not upgraded, that's the thing

MarinPostma: only some are, manually

honzasp: Hmm, maybe that's the root of the problem that should be fixed? :D

If we need to support old clients, the server must handle `/`, and if we need to support old servers, the clients must handle `/`, so this would mean that we are stuck with `/` anyway and we don't need to migrate clients to `/queries` :shrug:

However, I understand that we are laboring under constraints not under our control, so if we absolutely need to change the client behavior based on server version, I still think that we should refactor this function to return the version instead of the session, and use it only when we need the version.

honzasp (approved):

honzasp:
bors merge

PR #231, opened 2023-02-17
Hrana server

honzasp:
Implement a basic Hrana server for sqld.

To keep the PR small, the following features will be done in follow-up PRs:
- Execute statements concurrently (right now, all requests are handled in order, but the protocol supports reordering)
- Implement JWT authentication, so that we can expose the protocol on Turso databases
- Add tests (I only tested this using tests from `packages/js/hrana-client`)

### DIFF
--- a/sqld/src/hrana/mod.rs
+++ b/sqld/src/hrana/mod.rs
@@ -1,8 +1,8 @@
+use crate::database::service::DbFactory;
+use anyhow::{Context as _, Result};
+use enclose::enclose;
 use std::net::SocketAddr;
 use std::sync::Arc;
-use anyhow::{Result, Context as _};
-use enclose::enclose;
-use crate::database::service::DbFactory;
 
 mod conn;
 mod proto;
@@ -15,7 +15,8 @@ struct Server {
 pub async fn serve(db_factory: Arc<dyn DbFactory>, bind_addr: SocketAddr) -> Result<()> {
     let server = Arc::new(Server { db_factory });
 
-    let listener = tokio::net::TcpListener::bind(bind_addr).await
+    let listener = tokio::net::TcpListener::bind(bind_addr)
+        .await
         .context("Could not bind TCP listener")?;
     let local_addr = listener.local_addr()?;
     tracing::info!("Listening for Hrana connections on {}", local_addr);
@@ -33,7 +34,7 @@ pub async fn serve(db_factory: Arc<dyn DbFactory>, bind_addr: SocketAddr) -> Res
                     match conn::handle_conn(server, socket, conn_id).await {
                         Ok(_) => tracing::info!("Connection #{} was terminated", conn_id),
                         Err(err) => tracing::error!("Connection #{} failed: {:?}", conn_id, err),
-                    } 
+                    }
                 }});
 
                 conn_id += 1;
--- a/sqld/src/hrana/mod.rs
+++ b/sqld/src/hrana/mod.rs
@@ -1,0 +1,46 @@
+use std::net::SocketAddr;
+use std::sync::Arc;
+use anyhow::{Result, Context as _};
+use enclose::enclose;
+use crate::database::service::DbFactory;
+
+mod conn;
+mod proto;
+mod session;
+
+struct Server {
+    db_factory: Arc<dyn DbFactory>,
+}
+
+pub async fn serve(db_factory: Arc<dyn DbFactory>, bind_addr: SocketAddr) -> Result<()> {
+    let server = Arc::new(Server { db_factory });
+
+    let listener = tokio::net::TcpListener::bind(bind_addr).await
+        .context("Could not bind TCP listener")?;
+    let local_addr = listener.local_addr()?;
+    tracing::info!("Listening for Hrana connections on {}", local_addr);
+
+    let mut join_set = tokio::task::JoinSet::new();
+    let mut conn_id = 0;
+    loop {
+        tokio::select! {
+            accept_res = listener.accept() => {
+                let (socket, peer_addr) = accept_res
+                    .context("Could not accept a TCP connection")?;
+                tracing::info!("Accepted connection #{} from {}", conn_id, peer_addr);
+
+                join_set.spawn(enclose!{(server, conn_id) async move {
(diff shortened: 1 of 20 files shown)


sqld/src/hrana/session.rs
```diff
@@ -0,0 +1,193 @@
...
+
+fn proto_value_to_value(proto_value: proto::Value) -> Value {
+    match proto_value {
+        proto::Value::Null => Value::Null,
+        proto::Value::Integer { value } => Value::Integer(value),
+        proto::Value::Float { value } => Value::Real(value),
+        proto::Value::Text { value } => Value::Text(value),
+        proto::Value::Blob { value } => Value::Blob(value),
+    }
+}
```
MarinPostma, line 157: would implementing `From` here make sense?

sqld/src/hrana/session.rs
```diff
@@ -0,0 +1,193 @@
...
+                offset,
+                ..
+            } => ResponseError::SqlInputError {
+                source: sqlite_error,
+                message,
+                offset,
+            },
+            rusqlite_error => return Err(Error::RusqliteError(rusqlite_error)),
+        },
+        error => return Err(error),
```
MarinPostma, line 191: same here

sqld/src/main.rs
```diff
@@ -18,6 +18,9 @@ struct Cli {
     /// The address and port the PostgreSQL over WebSocket server listens to.
     #[clap(long, short, env = "SQLD_WS_LISTEN_ADDR")]
     ws_listen_addr: Option<SocketAddr>,
+    /// The address and port the Hrana server listens to.
+    #[clap(long, short = 'l', env = "SQLD_HRANA_LISTEN_ADDR")]
```
MarinPostma, line 22: I'm realizing that most shorthands don't make much sense, but I would keep the `-l` for HTTP, wdyt?

sqld/src/query_analysis.rs
```diff
@@ -95,17 +98,23 @@ impl Statement {
...
     }
 
     pub fn parse(s: &str) -> impl Iterator<Item = Result<Self>> + '_ {
         fn parse_inner(c: Cmd) -> Result<Statement> {
             let kind =
                 StmtKind::kind(&c).ok_or_else(|| anyhow::anyhow!("unsupported statement"))?;
+            let is_iud = matches!(
+                c,
+                Cmd::Stmt(Stmt::Insert { .. } | Stmt::Update { .. } | Stmt::Delete { .. })
+            );
```
MarinPostma, line 112: It feels like this should be embedded in StmtKind. Maybe it lacks granularity and should be extended

sqld/src/hrana/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+                    match conn::handle_conn(server, socket, conn_id).await {
+                        Ok(_) => tracing::info!("Connection #{} was terminated", conn_id),
+                        Err(err) => tracing::error!("Connection #{} failed: {:?}", conn_id, err),
+                    }
+                }});
+
+                conn_id += 1;
+            },
+            Some(task_res) = join_set.join_next() => {
+                task_res.expect("Hrana connection task failed")
```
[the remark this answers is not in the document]
honzasp: I tend to prefer to propagate panics, because a panic always means that the code contains a bug, so the safest way to proceed is to terminate the program. It also makes it more likely that we will notice (crashes are more visible than error logs).

However, I don't have a strong opinion about this, so if you prefer that the process survives panics, I can change the behavior.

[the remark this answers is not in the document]
MarinPostma: That makes sense, idk shared... Maybe it's safer to abort since there could be some invalid state...

sqld/src/hrana/session.rs
honzasp: I strongly prefer explicit functions to overloading `From`. Overloads make it harder to follow what is going on, especially when we are juggling multiple types with very similar names (such as `Value` and `proto::Value` here). `proto_value_to_value(value)` is much more descriptive than `value.into()`.

(In general, I tend to optimize for simplicity, and a plain function is simpler than an overloaded one.)

sqld/src/main.rs
honzasp: To be honest, I sneakily stole the `-l` shorthand in the hope that Hrana will become the default protocol, with HTTP being used only in very limited environments where WebSockets don't make sense :)

sqld/src/query_analysis.rs
honzasp: Yeah, I am a bit uneasy about this. I didn't want to modify `StmtKind` to avoid interfering with the query analysis code; the only reason to distinguish `INSERT`, `UPDATE` and `DELETE` from other writes like `ALTER TABLE` is to track the `affected_row_count`. But if you think that's ok, maybe I could split `StmtKind::Write` to `StmtKind::WriteRows` (for IUD) and `StmtKind::WriteOther`?

MarinPostma: let's go with that solution for now, and think about what to do with `StmtKind` later

sqld/src/hrana/session.rs
MarinPostma: I understand your POV, yet I also like to optimize for the principle of least astonishment. When you want to know if some type can be converted to another type in Rust, the usual thing you go look for is a `From` implementation. If you don't find it, you'll just assume that it doesn't exist and this creates confusion. Moreover, `From` creates an explicit link between two types (since the other type gets the reciprocal `Into` implementation for free).

I tend to agree that using `into()` makes it somewhat harder to follow what's going on, but its also because it usually don't really matter, but I find it hard to argue that `SomeType::from(something)` affects comprehension in any ways?

honzasp: Ok, you are here at home, let's use a trait :)

sqld/src/hrana/session.rs
honzasp: Ok. Though I have to say, I didn't get this signature right on the first try :)
```rust
impl TryFrom<Error> for ResponseError {
    type Error = Error;
    fn try_from(error: Error) -> Result<ResponseError, Error>;
}
```

PR #235, opened 2023-02-17
Hrana server authentication

honzasp:
Implement authentication in Hrana using JWTs. An Ed25519 public key is passed to sqld using `--hrana-jwt-key-file`, and the received JWT is verified using this key. We ignore the claims, except by validating the expiration in `exp`.

I also added a helper script `scripts/gen_jwt.py` to generate the key and a JWT.

This PR is stacked on top of #231.

### DIFF
--- a/sqld/src/hrana/mod.rs
+++ b/sqld/src/hrana/mod.rs
@@ -1,7 +1,9 @@
 use crate::database::service::DbFactory;
-use anyhow::{Context as _, Result};
+use anyhow::{Context as _, Result, bail};
 use enclose::enclose;
+use std::{fs, str};
 use std::net::SocketAddr;
+use std::path::Path;
 use std::sync::Arc;
 
 mod conn;
@@ -10,10 +12,15 @@ mod session;
 
 struct Server {
     db_factory: Arc<dyn DbFactory>,
+    jwt_key: Option<jsonwebtoken::DecodingKey>,
 }
 
-pub async fn serve(db_factory: Arc<dyn DbFactory>, bind_addr: SocketAddr) -> Result<()> {
-    let server = Arc::new(Server { db_factory });
+pub async fn serve(
+    db_factory: Arc<dyn DbFactory>,
+    bind_addr: SocketAddr,
+    jwt_key: Option<jsonwebtoken::DecodingKey>,
+) -> Result<()> {
+    let server = Arc::new(Server { db_factory, jwt_key });
 
     let listener = tokio::net::TcpListener::bind(bind_addr)
         .await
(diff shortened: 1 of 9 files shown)


sqld/src/hrana/mod.rs
```diff
@@ -45,3 +52,20 @@ pub async fn serve(db_factory: Arc<dyn DbFactory>, bind_addr: SocketAddr) -> Res
...
+        bail!("Received a private key, but a public key is expected")
+    } else if data.starts_with(b"-----BEGIN") {
+        bail!("Key is in unsupported PEM format")
+    } else if let Ok(data_str) = str::from_utf8(&data) {
+        jsonwebtoken::DecodingKey::from_ed_components(&data_str)
+            .context("Could not decode Ed25519 public key from base64")
+    } else {
+        bail!("Key is in an unsupported binary format")
+    }
+}
```
MarinPostma, line 71: This code would certainly be useful in auth!

sqld/src/hrana/session.rs
```diff
@@ -46,13 +54,27 @@ pub enum ResponseError {
...
     Ok(Session {
         streams: HashMap::new(),
     })
 }
 
+fn validate_jwt(jwt_key: &jsonwebtoken::DecodingKey, jwt: &str) -> Result<()> {
+    let validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::EdDSA);
+    match jsonwebtoken::decode::<serde_json::Value>(jwt, jwt_key, &validation) {
+        Ok(_token) => Ok(()),
+        Err(error) => bail!(ResponseError::AuthJwtRejected { source: dbg!(error) }),
```
MarinPostma, line 74:
```suggestion
        Err(error) => bail!(ResponseError::AuthJwtRejected { source: error }),
```

honzasp: Thanks, _of course_ I forgot to remove this :D

sqld/src/hrana/mod.rs
honzasp: The code in `http/auth.rs` seems to implement a diferent mechanism, instead of passing an argument pointing to a file with Ed25519 pubkey, we pass an argument that directly contains a base64-encoded RSA pubkey in PKCS#8 (which is awkwardly decoded by wrapping it in PEM headers and parsing it as PEM). However, the JWT auth in HTTP is not yet used in Turso, so perhaps we can use the Hrana mechanism also in HTTP?

MarinPostma: See clerk: https://clerk.dev/docs/request-authentication/validate-session-tokens?utm_source=www.google.fr&utm_medium=referral&utm_campaign=none

honzasp: We do a similar thing with TLS certificates and keys for gRPC: they are passed as env variables to the container and the entrypoint script stores them in temporary files and passes the paths to these files to sqld via the command line arguments.

Now that I think about it again, this is rather silly: it would be better if all keys/certs were passed to sqld in one of two ways:

- With a command line argument pointing to a file (useful when you run sqld standalone)
- With an env variable containing the data (useful when you run sqld in a container)

However, I'm not sure how we could implement this scheme with clap?

MarinPostma: Maybe @CodingDoug and @athoscouto input would be useful here

honzasp: I added the option of passing the key directly in env SQLD_HRANA_JWT_KEY, so we won't need to create temporary files in a container.

MarinPostma (approved):

honzasp:
bors merge

honzasp:
bors merge

PR #236, opened 2023-02-17
Transactional batches

MarinPostma:
This PR implemements transactional batches.

Before, when a statement failed in an HTTP batch, an error was returned, and we carried on processing the batch.

This PR makes is so that when a statement fails in a batch, a top-level error is returned and any ongoind transaction is rolledback.

This also fixes a pretty serious bug that could leave a transaction open on a shared http connection.

This change is breaking since top-level errors are now returned when a batch fails.


MarinPostma:
marking as draft for now, I want to add some tests, but open to discussion.
@CodingDoug tagging you specifically since you asked for this some time ago.

CodingDoug:
If I understand `is_transactional` correctly, this is checking to see if a statement is between a `begin` and `end`, and performs the rollback if there is an error.  But I can't tell how the caller would detect this (and the root cause) in the response other than by looking for a specific string message.  The following would make the response more actionable:

- A specific code for each top-level error case that won't change
- An indication of which statement caused the failure

MarinPostma:
> If I understand is_transactional correctly, this is checking to see if a statement is between a begin and end, and performs the rollback if there is an error.

Transactional means to things:
- any failure in the statement batch should cause a rollback
- after the execution of the batch, the connection should not be in a transaction state, i.e the connection must be left in a clean state.

Error handling still needs more work to handle reporting errors as I would like to, especially at the primary/replica interface, so that aspect is still pretty much WIP

MarinPostma:
converting to draft, this needs a bit more work.

MarinPostma:
closed in favor of #307

PR #239, opened 2023-02-18
Make named parameters work with :name, $name, @name and name

### LINKED ISSUE #237: Named parameters don't work
Table schema
```
CREATE TABLE IF NOT EXISTS counter(country TEXT, city TEXT, value INT, PRIMARY KEY(country, city)) WITHOUT ROWID
```
Insert/Update
```
curl -X POST -d '{"statements":[{"q":"insert into counter(country, city, value) values(\"US\", :city, 1)","params":{"city":"GDA"}}]}' https://haaawk:<password>@haaawk1-haaawk.turso.io
[{"error":{"message":"NOT NULL constraint failed: counter.city"}}]%
```
Same SQL with positional parameter works:
```
curl -X POST -d '{"statements":[{"q":"insert into counter(country, city, value) values(\"US\", ?, 1)","params":["GDA"]}]}' https://haaawk:<password>@haaawk1-haaawk.turso.io
[{"results":{"columns":[],"rows":[]}}]%
```
It seems that when using parameter name `:city` instead of `city` in params, everything works:
```
curl -X POST -d '{"statements":[{"q":"insert into counter(country, city, value) values(\"US\", :city, 1)","params":{":city":"NYC"}}]}' https://haaawk:<password>@haaawk1-haaawk.turso.io
[{"results":{"columns":[],"rows":[]}}]%
```
Same when naming parameter `@city` instead of `:city`:
```
% curl -X POST -d '{"statements":[{"q":"insert into counter(country, city, value) values(\"US\", $city, 1)","params":{"$city":"SEA"}}]}' https://haaawk:<password>@haaawk1-haaawk.turso.io
[{"results":{"columns":[],"rows":[]}}]%
```

haaawk:
Fixes #237

### DIFF
--- a/sqld/src/query.rs
+++ b/sqld/src/query.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::convert::Infallible;
 use std::str::FromStr;
 
-use anyhow::{ensure, Context};
+use anyhow::{anyhow, ensure, Context};
 use futures::stream;
 use pgwire::api::results::{text_query_response, FieldInfo, Response, TextDataRowEncoder};
 use pgwire::api::Type as PgType;
@@ -318,11 +318,6 @@ impl Params {
     pub fn bind(&self, stmt: &mut rusqlite::Statement) -> anyhow::Result<()> {
         let param_count = stmt.parameter_count();
         ensure!(
-            param_count <= self.len(),
-            "missing parameters, expected {param_count} found {}",
-            self.len()
-        );
-        ensure!(
             param_count >= self.len(),
             "too many parameters, expected {param_count} found {}",
             self.len()
@@ -330,9 +325,11 @@ impl Params {
 
         if param_count > 0 {
             for index in 1..=param_count {
+                let mut param_name = None;
                 // get by name
                 let maybe_value = match stmt.parameter_name(index) {
                     Some(name) => {
+                        param_name = Some(name);
                         let mut chars = name.chars();
                         match chars.next() {
                             Some('?') => {
@@ -341,7 +338,9 @@ impl Params {
                                 )?;
                                 self.get_pos(pos)
                             }
-                            _ => self.get_named(name),
+                            _ => self
+                                .get_named(name)
+                                .or_else(|| self.get_named(chars.as_str())),
                         }
                     }
                     None => self.get_pos(index),
@@ -349,6 +348,10 @@ impl Params {
 
                 if let Some(value) = maybe_value {
                     stmt.raw_bind_parameter(index, value)?;
+                } else if let Some(name) = param_name {
+                    return Err(anyhow!("value for parameter {} not found", name));
+                } else {
+                    return Err(anyhow!("value for parameter {} not found", index));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


sqld/src/query.rs
```diff
@@ -381,8 +381,8 @@ mod test {
         let con = rusqlite::Connection::open_in_memory().unwrap();
         let mut stmt = con.prepare("SELECT :first || $second").unwrap();
         let mut params = HashMap::new();
-        params.insert(":first".to_owned(), Value::Integer(10));
-        params.insert("$second".to_owned(), Value::Integer(20));
+        params.insert("first".to_owned(), Value::Integer(10));
+        params.insert("second".to_owned(), Value::Integer(20));
```
haaawk, line 385: 😄

sqld/src/query.rs
```diff
@@ -389,6 +391,19 @@ mod test {
         assert_eq!(stmt.expanded_sql().unwrap(), "SELECT 10 || 20");
     }
 
+    #[test]
+    fn test_bind_params_positional_named_no_prefix() {
+        let con = rusqlite::Connection::open_in_memory().unwrap();
+        let mut stmt = con.prepare("SELECT :first || $second").unwrap();
```
MarinPostma, line 397: What happens if you try with a statement like 'select :first || $first'?

haaawk: It won't allow it:
```
missing parameters, expected 2 found 1
```
here's the change:
```
diff --git a/sqld/src/query.rs b/sqld/src/query.rs
index b75bdf5..739f689 100644
--- a/sqld/src/query.rs
+++ b/sqld/src/query.rs
@@ -394,10 +394,9 @@ mod test {
     #[test]
     fn test_bind_params_positional_named_no_prefix() {
         let con = rusqlite::Connection::open_in_memory().unwrap();
-        let mut stmt = con.prepare("SELECT :first || $second").unwrap();
+        let mut stmt = con.prepare("SELECT :first || $first").unwrap();
         let mut params = HashMap::new();
         params.insert("first".to_owned(), Value::Integer(10));
-        params.insert("second".to_owned(), Value::Integer(20));
         let params = Params::new_named(params);
         params.bind(&mut stmt).unwrap();
```

haaawk: When I worked around it by adding second parameter `$first`, it failed:
```
```
here's the change:
```
```

haaawk: Thanks for pointing it out @MarinPostma

haaawk:
I swapped the order of fetching named params so that $name, :name and @name will take precedence over name.
Added a test to verify that.

MarinPostma (commented):
can you also update this document to reflect the changes? https://github.com/libsql/sqld/blob/main/docs/http_api.md

sqld/src/query.rs
haaawk: done.

haaawk:
I rebased, changed the check that number of params in statement <= number of params in request to checking that each param in statement has a match in request and updated the docs.

MarinPostma (approved):
looks good :)

bors merge

MarinPostma:
bors merge

PR #241, opened 2023-02-21
system monitor

MarinPostma:
I got a bit carried away experimenting with graceful shutdown for primary.

I'll leave this open as draft while collecting, but it's maybe a bit too complex in the current shape.


MarinPostma:
I think I'll extract that to a library and polish the API before integrating it.

PR #242, opened 2023-02-21
primary idle shutdown

MarinPostma:
This PR adds idle shutdown to the primary by sharing the IdleShutdownService with the rpc server.

Whenever the primary does not receive any ping from replicas for some period *AND* doesn't get any requests, then it shuts down.


haaawk (approved):
Amazing! Thanks a lot @MarinPostma

haaawk:
I think we just need to fix clippy and we're good to go. Let's get this merged today and I can test it and enable it on Turso tomorrow.

MarinPostma:
bors merge

PR #244, opened 2023-02-21
Fix assertion in replication client

haaawk:


sqld/src/replication/client.rs
```diff
@@ -434,6 +434,7 @@ impl ReadReplicationHook {
     /// Asks the writer for new log frames to apply.
     async fn fetch_log_entries(&mut self) -> anyhow::Result<()> {
         let current_offset = self.current_frame_index();
+        dbg!(current_offset);
```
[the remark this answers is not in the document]
haaawk: Yeah. I will remove that. But it's not a problem because I actually made a mistake in a fix.

sqld/src/replication/client.rs
```diff
@@ -445,7 +446,7 @@ impl ReadReplicationHook {
                     let frame = WalFrame::decode(raw_frame.data)?;
                     debug_assert_eq!(
                         Some(frame.header.frame_id),
-                        self.wal_index_meta.map(|m| m.pre_commit_index),
+                        Some(current_offset.unwrap_or(0) + frame_count + 1),
```
haaawk, line 449: After some more thoughts I thing None should translate to -1 so it should be something like `current_offset.map(|x| x + frame_count + 1).or(Some(0)),`

haaawk: Or rather `current_offset.map(|x| x + frame_count + 1).or(Some(frame_count)),`

MarinPostma (approved):
bors merge

PR #247, opened 2023-02-22
Unify HTTP and Hrana authentication

honzasp:
Unify authentication in the HTTP API and in Hrana into a single `Auth` object. This keeps the legacy HTTP Basic auth, but it uses the same JWTs as Hrana for HTTP Bearer auth.

To specify the legacy Basic auth, we can continue using the `--http-auth` parameter with `basic:$PARAM` as before. However, to enable both Bearer authentication and Hrana authentication, we either pass a path to the JWT key in `--auth-jwt-key-file` or we pass the key directly in env `SQLD_AUTH_JWT_KEY`.

As a side effect of this PR, when HTTP Basic auth is specified but the JWT key is not set, Hrana server will not accept any connections, because Hrana only supports JWT authentication.


sqld/src/auth.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[error("The JWT has expired")]
+    JwtExpired,
+    #[error("The JWT is immature (not valid yet)")]
+    JwtImmature,
+    #[error("Authentication failed")]
+    Other,
+}
+
+/// A witness that the user has been authenticated.
+pub struct Authenticated(());
```
MarinPostma, line 41: question: what's the difference between `Authenticated` and `Authenticated(())`?

MarinPostma (approved):

sqld/src/auth.rs
honzasp: `Authenticated(())` is a tuple struct with a single private field of type `()`. The point of this is that the value can only be created from this module, so it can act as a witness/proof that the user has been authenticated. This approach is inspired from rustls:
https://docs.rs/rustls/latest/rustls/manual/_01_impl_vulnerabilities/index.html#goto-fail

MarinPostma: interesting trick :)

honzasp:
bors merge

PR #251, opened 2023-02-27
feat: update pgwire to 0.11

sunng87:
This patch updates pgwire to latest release by:

- Make QueryHandler per-connection instead of per-request, this allows it to cache prepared statements
- Add `include_column_defs` to `ResultSet` enables control of whether to output row descriptions
- Adopt API changes

On prepared statement, aka extended query in postgresql:

- The current implementation has support for parse-bind-describe-execute process, which is used by most postgresql clients like jdbc and psycopg, and it should work for sqld at the moment
- Some clients sends `describe` on statement after `parse` to inference parameters types and resultset column types, this is not work on sqld because it requires type inference without execution. Even sqlite has weak support for that, and to we will also need to update RPC for that.
- pgwire is getting closer to a relatively stable API for prepared statement.


MarinPostma:
thanks a lot for your contribution! I'll review this PR ASAP :)

haaawk:
> @MarinPostma The JWT test failure doesn't seem to be related to this pull request?

Yes. They are failing due to expired JWT token. @honzasp fixed that here https://github.com/libsql/sqld/pull/253. Probably rebasing this PR on top of main should help.

haaawk:
I restarted the job - maybe it will pick up the new main.

MarinPostma:
bors try

MarinPostma (commented):
@sunng87 Why is that ok to not respond to a `describe` message? won't the client be expecting a response? Why not just always include column descriptions in the response as we did before since we can't really handle `describe` anyways?

sunng87:
@MarinPostma Good question! 

As I explained, there are two kinds of `Describe`, on statement and on portal, both are issued before `Execute`. So in an implementation that fully compatible with postgres, it should be able to inference types for results without actually executing it.

Here we only support `Describe` on portal, because in most case it's followed by an `Execute`. So we respond `Describe` in `do_query` with both row description and data rows. However, when clients cached the statement, it won't issue `Describe` before `Execute`. I use `include_column_defs` to mark that.

For the other type of `Describe` issued on statement, clients want server to inference parameter types and row description without executing it. I'm afraid sqlite has just limited support for that (column.get_decl_type iirc) . So in sqld we don't respond it for now.

penberg (approved):

PR #253, opened 2023-02-27
Replace an expired JWT in tests

honzasp:
The previous JWT expired in 3 days, breaking tests. The new one should expire in 2123, when we should no longer be around!


haaawk (approved):

honzasp:
bors merge

PR #262, opened 2023-03-02
sqld: Disable Tokio signal support

penberg:
Unikraft doesn't support signals, but we don't really even need them, so disable them for Tokio.


psarna (approved):

penberg:
bors merge

PR #272, opened 2023-03-07
Upgrade Hrana WebSocket connections from HTTP

honzasp:
Our infrastructure on Fly.io can only handle connections on port 443, so we need to use the same port for HTTP and for Hrana. Fortunately, WebSockets were specifically designed to allow this! The WebSocket handshake is an HTTP request with `Connection: upgrade` and `Upgrade: websocket` headers, so we can just hijack these HTTP requests and pass them to Hrana.

Note that Hrana still needs to be enabled using the `-l` flag, so it's currently impossible to enable Hrana without binding to a Hrana-specific port in addition to the HTTP port.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -603,9 +603,9 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.63.0"
+version = "0.64.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


MarinPostma (approved):

honzasp:
bors merge

haaawk:
@honzasp please fix the formatting

honzasp:
There is nothing wrong with formatting, the check failed because it could not install Rust.
bors retry

PR #273, opened 2023-03-07
Kick the `IdleShutdownLayer` in Hrana server

honzasp:
Stacked on top of #272

### DIFF
--- a/sqld/src/hrana/mod.rs
+++ b/sqld/src/hrana/mod.rs
@@ -1,6 +1,6 @@
 use crate::auth::Auth;
 use crate::database::service::DbFactory;
-use crate::utils::services::idle_shutdown::IdleShutdownLayer;
+use crate::utils::services::idle_shutdown::IdleKicker;
 use anyhow::{Context as _, Result};
 use enclose::enclose;
 use std::net::SocketAddr;
@@ -16,7 +16,7 @@ mod session;
 struct Server {
     db_factory: Arc<dyn DbFactory>,
     auth: Arc<Auth>,
-    idle_shutdown_layer: Option<IdleShutdownLayer>,
+    idle_kicker: Option<IdleKicker>,
     next_conn_id: AtomicU64,
 }
 
@@ -29,14 +29,14 @@ pub struct Upgrade {
 pub async fn serve(
     db_factory: Arc<dyn DbFactory>,
     auth: Arc<Auth>,
-    idle_shutdown_layer: Option<IdleShutdownLayer>,
+    idle_kicker: Option<IdleKicker>,
     bind_addr: SocketAddr,
     mut upgrade_rx: mpsc::Receiver<Upgrade>,
 ) -> Result<()> {
     let server = Arc::new(Server {
         db_factory,
         auth,
-        idle_shutdown_layer,
+        idle_kicker,
         next_conn_id: AtomicU64::new(0),
     });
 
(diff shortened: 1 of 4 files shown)


sqld/src/hrana/mod.rs
```diff
@@ -44,6 +48,10 @@ pub async fn serve(
 
     let mut join_set = tokio::task::JoinSet::new();
     loop {
+        if let Some(isl) = server.idle_shutdown_layer.as_ref() {
+            isl.kick();
+        }
+
```
haaawk, line 54: would it make sense to do `server.idle_shutdown_layer.as_ref().map(|s| s.kick());` instead?

sqld/src/utils/services/idle_shutdown.rs
```diff
@@ -37,6 +37,10 @@ impl IdleShutdownLayer {
             watcher: Arc::new(sender),
         }
     }
+
+    pub fn kick(&self) {
+        let _: Result<_, _> = self.watcher.send(());
```
haaawk, line 42: Why is it ok to ignore this result?

haaawk: It seems that `let _ = self.watcher.send(());` should be enough, no?

haaawk (commented):
LGTM in general

sqld/src/utils/services/idle_shutdown.rs
honzasp: Yes, it would be enough, but the explicit type makes sure I actually drop the type that I want. It has happened to me that I wrote `let _ = channel.send(something);` where `channel` was a mpsc channel where the `send()` method is async, so instead of sending an item on the channel and ignoring the result, I ignored the future! This taught me to specify the ignored type.

By a funny coincidence, I made the same mistake today, and the compiler caught the error because I specified the type.

sqld/src/utils/services/idle_shutdown.rs
```diff
@@ -37,6 +37,10 @@ impl IdleShutdownLayer {
             watcher: Arc::new(sender),
         }
     }
+
+    pub fn kick(&self) {
```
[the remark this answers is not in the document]
honzasp: Hmm, `IdleShutdownLayer` contains a `Sender`, it is initialized from a `sender`, and the only thing it does is to pass the `watcher` (i.e. `sender`) to the `IdleShutdownService`, which does `self.watcher.send()`. It surely looks like a sender to me!

Maybe I should add a `IdleKicker` type, that would be isomorphic to the `IdleShutdownLayer`, but would be used to kick the watchdog instead of being used as a Tower layer?

sqld/src/hrana/mod.rs
honzasp: Unfortunately, the uglifier transforms this into

```rust
        conn.server
            .idle_shutdown_layer
            .as_ref()
            .map(|isl| isl.kick());
```

which is both longer and not more readable than the explicit version :(

haaawk: I like the uglifier name :D

sqld/src/utils/services/idle_shutdown.rs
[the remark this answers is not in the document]
honzasp: Done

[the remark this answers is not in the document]
MarinPostma: You're right I read a bit to quickly, this is a sender, but yeah isomorphic type is better, so the layer remains for use in hyper related stuff

honzasp:
bors merge

honzasp:
bors merge

PR #276, opened 2023-03-09
Add `last_insert_rowid` to Hrana statement result

honzasp:
This information may be useful, and it's required to implement some client libraries (for example, to emulate the Node.js `sqlite3` package).


haaawk:
Golang driver needs that too

haaawk (approved):
LGTM

PR #281, opened 2023-03-10
load dumps from CLI

MarinPostma:
This PR enables loading a dump file from the CLI.

### DIFF
--- a/sqld/src/database/dump_loader.rs
+++ b/sqld/src/database/dump_loader.rs
@@ -1,0 +1,91 @@
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::path::PathBuf;
+
+use anyhow::anyhow;
+use sqld_libsql_bindings::wal_hook::WalHook;
+use tokio::sync::{mpsc, oneshot};
+
+use super::libsql::open_db;
+
+type OpMsg = Box<dyn FnOnce(&rusqlite::Connection) + 'static + Send + Sync>;
+
+pub struct DumpLoader {
+    sender: mpsc::Sender<OpMsg>,
+}
+
+impl DumpLoader {
+    pub async fn new(
+        path: PathBuf,
+        hooks: impl WalHook + Clone + Send + 'static,
+    ) -> anyhow::Result<Self> {
+        let (sender, mut receiver) = mpsc::channel::<OpMsg>(1);
+
+        let (ok_snd, ok_rcv) = oneshot::channel::<anyhow::Result<()>>();
+        tokio::task::spawn_blocking(move || {
+            let db = match open_db(path, hooks, false) {
+                Ok(db) => {
+                    let _ = ok_snd.send(Ok(()));
+                    db
+                }
+                Err(e) => {
+                    let _ = ok_snd.send(Err(e));
+                    return;
+                }
+            };
+            while let Some(f) = receiver.blocking_recv() {
+                f(&db);
+            }
+        });
+
+        ok_rcv.await??;
+
+        Ok(Self { sender })
+    }
+
+    /// Attempts to load the dump at `path` into the database.
+    pub async fn load_dump(&self, path: PathBuf) -> anyhow::Result<()> {
+        tracing::info!("loading dump at `{}`", path.display());
+        let (snd, ret) = oneshot::channel();
+        self.sender
+            .send(Box::new(move |conn| {
+                let ret = perform_dump(conn, path);
+                let _ = snd.send(ret);
+            }))
+            .await
+            .map_err(|_| anyhow!("dump loader channel closed"))?;
+
+        ret.await??;
+
+        tracing::info!("dump loaded sucessfully");
+
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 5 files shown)


sqld/src/database/dump_loader.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub async fn load_dump(&self, path: PathBuf) -> anyhow::Result<()> {
+        tracing::info!("loading dump at `{}`", path.display());
+        let (snd, ret) = oneshot::channel();
+        self.sender
+            .send(Box::new(move |conn| {
+                let ret = perform_dump(conn, path);
+                let _ = snd.send(ret);
+            }))
+            .await
+            .map_err(|_| anyhow!("dump loader channel closed"))?;
```
athoscouto, line 56: If the dump loading fails, will this message be all we see?
If that is the case, can we expose the underlying error with a message more meaningful to users?
Something like `loading DB from dump failed: <underlying error here>`.

sqld/src/database/dump_loader.rs
```diff
@@ -0,0 +1,91 @@
...
+                    db
+                }
+                Err(e) => {
+                    let _ = ok_snd.send(Err(e));
+                    return;
+                }
+            };
+            while let Some(f) = receiver.blocking_recv() {
+                f(&db);
+            }
```
athoscouto, line 38: Just out of curiosity, to check if I got things right.
This will be executed only once, right?

sqld/src/lib.rs
```diff
@@ -213,9 +215,17 @@ async fn start_replica(
     let enable_bottomless = config.enable_bottomless_replication;
     #[cfg(not(feature = "bottomless"))]
     let enable_bottomless = false;
+    let hook = ReplicationLoggerHook::new(logger.clone());
+
+    // load dump is necessary
+    let dump_loader = DumpLoader::new(config.db_path.clone(), hook.clone()).await?;
+    if let Some(ref path) = config.load_dump_path {
+        dump_loader.load_dump(path.into()).await?;
+    }
```
athoscouto, line 224: Why only on `start_primary`?

athoscouto (commented):
Should we prevent loading from a dump in case there is any content on the current DB?

sqld/src/main.rs
```diff
@@ -105,6 +105,11 @@ struct Cli {
     /// By default, the server doesn't shutdown when idle.
     #[clap(long, env = "SQLD_IDLE_SHUTDOWN_TIMEOUT_S")]
     idle_shutdown_timeout_s: Option<u64>,
+
+    /// Load the dump at the provided path.
+    /// Requires that the node is not in replica mode
+    #[clap(long, env = "SQLD_LOAD_DUMP_PATH", conflicts_with = "primary_grpc_url")]
+    load_dump_path: Option<PathBuf>,
```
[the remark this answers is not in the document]
MarinPostma: I reserve `dump_path` for  the path at which dumps would be performed, as opposed to loaded, `load_from_dump` is fine

sqld/src/main.rs
```diff
@@ -105,6 +105,11 @@ struct Cli {
     /// By default, the server doesn't shutdown when idle.
     #[clap(long, env = "SQLD_IDLE_SHUTDOWN_TIMEOUT_S")]
     idle_shutdown_timeout_s: Option<u64>,
+
+    /// Load the dump at the provided path.
+    /// Requires that the node is not in replica mode
+    #[clap(long, env = "SQLD_LOAD_DUMP_PATH", conflicts_with = "primary_grpc_url")]
```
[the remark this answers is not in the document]
MarinPostma: because a replica cannot load a dump, only the primary can

sqld/src/lib.rs
MarinPostma: If a replica load a dump, then there is no guarantee it's state is consistent with the primary. Hence, loading on a replica is forbidden

MarinPostma:
> Should we prevent loading from a dump in case there is any content on the current DB?

I want to potentially reuse that code for HTTP later on. There is a difference between loading a dump and loading from a dump. You could, in theory, load a dump to an already running database, and if there are no conflicts this is ok. when loading *from* a dump, you need to start from a fresh state.

From experience, a destructive dump load is not a good solution. When loading *from* a dump, I can add a check that refuses to load it if there is something there already though

sqld/src/database/dump_loader.rs
MarinPostma: once for every call to load dump

sqld/src/database/dump_loader.rs
MarinPostma: this error is unreachable in reality, the only exit condition to the loop is that we dropped the last sender, but we're currently holding one. I can't unwrap because the closure does not implement debug, so I just mock returning an error

MarinPostma:
@athoscouto I have changed the CLI param name, and added an error when trying to load from a dump and a database already exists.

athoscouto (approved):
looks good

MarinPostma:
bors merge

PR #290, opened 2023-03-13
implement 2 phase commit for logging frames to replication log

MarinPostma:
This PR implements a 2 phase commit between the actual WAL and the replication log, so no operation end in the replication log if they were not successfully committed to the database before.


docker-entrypoint.sh
```diff
@@ -8,7 +8,7 @@ SQLD_DB_PATH="${SQLD_DB_PATH:-iku.db}"
 SQLD_PG_LISTEN_ADDR="${SQLD_PG_LISTEN_ADDR:-"0.0.0.0:5432"}"
 SQLD_HTTP_LISTEN_ADDR="${SQLD_HTTP_LISTEN_ADDR:-"0.0.0.0:8080"}"
 
-SQLD_HTTP_AUTH="${SQLD_HTTP_AUTH:-"always"}"
+# SQLD_HTTP_AUTH="${SQLD_HTTP_AUTH:-"always"}"
```
penberg, line 11: leftovers from debugging?

docker-entrypoint.sh
```diff
@@ -21,7 +21,7 @@ if [ "$1" = '/bin/sqld' ]; then
 
   # Listen on HTTP 8080 port by default.
   server_args+=("--http-listen-addr" "$SQLD_HTTP_LISTEN_ADDR")
-  server_args+=("--http-auth" "$SQLD_HTTP_AUTH")
+  # server_args+=("--http-auth" "$SQLD_HTTP_AUTH")
```
penberg, line 24: ditto

docker-entrypoint.sh
MarinPostma: yeah, but actually what is that? I can't launch the server with this set to its default. what's that `always`?

MarinPostma:
bors merge

PR #294, opened 2023-03-15
Add conditional execution and compute requests to Hrana

honzasp:

- Execute a `BEGIN` statement
- ...

---

This PR changes that by introducing a mechanism for conditional execution of statements on the server. This mechanism is described in the Hrana spec, but a tl;dr is as follows:

- The server maintains a set of variables with values.
- A `compute` request can be used to execute operations that evaluate expressions and mutate the variables.
- The `execute` request can include a condition; if the condition evaluates to false, the statement is not executed. It can also contain a sequence of operations that will be executed when the statement was successful or when it failed.

This mechanism will allow clients to implement batched transactions using a single roundtrip. An alternative solution would be to bake the concept of a "batch" into the protocol, but it would have similar implementation complexity to this solution, while being much less general.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2185,6 +2185,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "348108ab3fba42ec82ff6e9564fc4ca0247bdccdc68dd8af9764bbc79c3c8ffb"
 
 [[package]]
+name = "libsql-client"
+version = "0.6.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


honzasp:
After discussion with Marin it turns out that this extension optimizes roundtrips between the client and replica, but we still have the roundtrips between replica and master. To make it possible to easily execute batches on the master, we decided to change the protocol to introduce the concept of a "program", which is a sequence of conditional executions and compute operations. Imporantly, variables are scoped to the program, so the program can be easily transferred from the master to the replica.

CodingDoug:
"Program" sounds a lot like a stored procedure. :)

honzasp:
That's not a coincidence :) The difference is that the program is not stored in the database, and in the initial version it is only a sequence of conditional steps (no loops).

PR #295, opened 2023-03-15
Hrana programs

honzasp:

- Execute a `BEGIN` statement
- ...

---

This PR changes that by introducing the concept of "program" to the Hrana protocol. A program is a sequence of SQL statements interleaved with (very rudimentary) computation. The program maintains state in variables and it can decide to skip some statements. It can also react to statements succeeding or failing.

This mechanism will allow clients to implement batched transactions using a single roundtrip. An alternative solution would be to bake the concept of a "batch" into the protocol, but it would have similar implementation complexity to this solution, while being much less general.

This PR replaces #294 after discussion with @MarinPostma. Wrapping batches into self-contained programs makes it much easier to send them from the replica for execution to the master.


honzasp:
After discussions with @MarinPostma, we decided to pursue yet another design: this approach is unnecessarily complex to implement batches, but too simplistic to implement general programs.

PR #298, opened 2023-03-22
limit replica frame buffer size

MarinPostma:
This PR limits the number of frames the replica can buffer when fetching from the primary

If a transaction exceeds the maximum buffer size (1000 pages), then the replica will fail to make progress. I will address this issue in a subsequent issue.


MarinPostma:
bors merge

PR #300, opened 2023-03-22
Hrana batches

honzasp:

- Execute a `BEGIN` statement
- ...

---

This PR changes that by introducing the concept of "batch" to the Hrana protocol. A batch is a sequence of steps, and each step is an SQL statement with optional condition. The condition can refer to success or failure of previous steps. This mechanism will allow clients to implement batched transactions using a single roundtrip.

This PR replaces #294 and #295 after discussion with @MarinPostma.


MarinPostma (approved):

honzasp:
bors merge

PR #301, opened 2023-03-22
Snapshot merging

MarinPostma:
This PR implements snapshot merging. When the size taken by snapshots grow larger than twice the size of the database, or when there are more than 32 snapshot files, then all snapshots are merged together.


MarinPostma (commented):
bors merge

PR #305, opened 2023-03-24
Add HTTP API v1 ("Hrana over HTTP")

honzasp:
We need to update the HTTP API to support batches and to fix some issues (such as encoding of types). Instead of coming up with a new API, let's just reuse the Hrana structures and semantics. In effect, the new version of the HTTP API is "Hrana over HTTP".

This is needed for [the new generation of the client libraries](https://github.com/libsql/libsql-client-ts/pull/10).

### DIFF
--- a/sqld/src/http/hrana_over_http.rs
+++ b/sqld/src/http/hrana_over_http.rs
@@ -1,0 +1,108 @@
+use anyhow::{anyhow, Context, Result};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::sync::Arc;
+
+use crate::database::service::DbFactory;
+use crate::{batch, hrana};
+
+#[derive(thiserror::Error, Debug)]
+enum ResponseError {
+    #[error("Could not parse request body: {source}")]
+    BadRequestBody { source: serde_json::Error },
+
+    #[error(transparent)]
+    Stmt(batch::StmtError),
+}
+
+pub async fn handle_index(
+    _req: hyper::Request<hyper::Body>,
+) -> Result<hyper::Response<hyper::Body>> {
+    let body = "This is sqld HTTP API v1 (\"Hrana over HTTP\")";
+    let body = hyper::Body::from(body);
+    Ok(hyper::Response::builder()
+        .header("content-type", "text/plain")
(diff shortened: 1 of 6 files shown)


MarinPostma (commented):
@honzasp 
I tried executing: `echo '{ "stmt": {"sql": "select * from test", "want_rows": true }}' | http post "127.0.0.1:8080/v1/execute"`

The returned error is a 500, and gets logged in the logs: 
```
2023-03-28T11:46:36.206182Z  INFO sqld::http: got request: POST /v1/execute
2023-03-28T11:46:36.206974Z ERROR tower_http::trace::on_failure: response failed classification=Status code: 500 Internal Server Error latency=0 ms
```

I am not sure this is adequate, and further error categorization should probably be performed

sqld/src/http/hrana_over_http.rs
```diff
@@ -0,0 +1,157 @@
...
+            StmtError::SqlParse { .. }
+            | StmtError::SqlNoStmt
+            | StmtError::SqlManyStmts
+            | StmtError::ArgsInvalid { .. }
+            | StmtError::SqlInputError { .. } => hyper::StatusCode::BAD_REQUEST,
+            StmtError::ArgsBothPositionalAndNamed => hyper::StatusCode::NOT_IMPLEMENTED,
+            StmtError::TransactionTimeout | StmtError::TransactionBusy => {
+                hyper::StatusCode::SERVICE_UNAVAILABLE
+            }
+            StmtError::SqliteError { .. } => hyper::StatusCode::INTERNAL_SERVER_ERROR,
```
MarinPostma, line 132: Uhm regarding my comment before: I think this is the culprit

MarinPostma: I was just surprised because it came up in the logs as an internal error, which leads me to believe that there is something broken, although everything is fine, and this is simply a user error

MarinPostma: not sure
let's start with that and work on the error story later

MarinPostma (approved):

MarinPostma:
bors merge

PR #307, opened 2023-03-27
sqld programs

MarinPostma:
This PR implements sqld programs on the model of hrana programs.

Programs take over batches in the execution of queries within sqld, since they are very flexible and let the `Database` API consumer specify how statements should be executed.

The API of the sqld programs closely matches that of hrana batches, so the implementation on hrana side is changed to simply transform a hrana batch into a sqld program.

On the HTTP side, batches of queries are transformed into a program that causes a rollback if any statement in the batch fails.

On the replica, the programs are serialized and sent over by gRPC to the primary when we detect that they could cause a write. 

This PR also solves a known bug in HTTP batches that used to unconditionally execute statements in a batch even if a statement in the batch failed, and never rolledback, since the new implementation of HTTP batches ensures that a rollback is always performed if any statement in the batch fails to execute.

Future work: expose a more direct access to the `Database` to the consumers


sqld/src/database/write_proxy.rs
```diff
@@ -118,39 +119,33 @@ impl WriteProxyDatabase {
...
 #[async_trait::async_trait]
 impl Database for WriteProxyDatabase {
-    async fn execute_batch(&self, queries: query::Queries) -> Result<(Vec<QueryResult>, State)> {
+    async fn execute_program(&self, pgm: Program) -> Result<(Vec<Option<QueryResult>>, State)> {
         let mut state = self.state.lock().await;
         if *state == State::Init
-            && queries.iter().all(|q| q.stmt.is_read_only())
-            && final_state(*state, queries.iter().map(|s| &s.stmt)) == State::Init
+            && pgm.is_read_only()
+            && final_state(*state, pgm.steps.iter().map(|s| &s.query.stmt)) == State::Init
```
honzasp, line 126: I'm not sure if this is correct: some statements in the program can be skipped, but `final_state()` assumes that we always execute all statements.

sqld/src/hrana/batch.rs
```diff
@@ -0,0 +1,89 @@
...
+};
+
+#[derive(thiserror::Error, Debug)]
+pub enum BatchError {
+    #[error("Invalid reference to step in a condition")]
+    CondBadStep,
+}
+
+fn proto_cond_to_cond(cond: &proto::BatchCond) -> Result<Cond> {
+    let try_convert_step = |step: i32| -> std::result::Result<usize, BatchError> {
```
honzasp, line 18: When we are already checking the step index, perhaps we can also check that it refers to step that is before the current step?

sqld/src/query_analysis.rs
```diff
@@ -70,7 +70,11 @@ pub enum State {
 impl State {
     pub fn step(&mut self, kind: StmtKind) {
         *self = match (*self, kind) {
-            (State::Txn, StmtKind::TxnBegin) | (State::Init, StmtKind::TxnEnd) => State::Invalid,
+            // those two transition will cause an error, but since we are interested in what the
+            // pesimistic final state is, and we will adjust when we get the actual state back.
+            (State::Txn, StmtKind::TxnBegin) => State::Txn,
+            (State::Init, StmtKind::TxnEnd) => State::Init,
```
honzasp, line 76: What does it mean that a state is pesimistic? Why not use `State::Invalid` as the result of these invalid transitions?

sqld/src/database/libsql.rs
```diff
@@ -266,11 +185,153 @@ impl LibSqlDb {
...
+    ) -> Option<QueryResult> {
+        let enabled = match step.cond.as_ref() {
+            Some(cond) => match eval_cond(cond, results) {
+                Ok(enabled) => enabled,
+                Err(e) => return Some(Err(e)),
+            },
+            None => true,
+        };
+
+        enabled.then(|| self.execute_query(&step.query))
```
honzasp, line 231: Nice! I didn't know about `bool::then()`, but I often thought that it would be useful :)

honzasp (commented):
Looks ok, just a few comments (one important about correctness).

sqld/src/database/mod.rs
```diff
@@ -9,17 +9,88 @@ pub mod write_proxy;
...
-    /// Executes a batch of queries, and returns a vec of results corresponding to the queries,
-    /// and the state the database is in after the call to execute.
-    async fn execute_batch(&self, queries: Queries) -> Result<(Vec<QueryResult>, State)>;
+    /// Execute all the queries in the batch sequentially.
+    /// If an query in the batch fails, the remaining queries are ignores, and the batch current
+    /// transaction (if any) is rolledback.
+    async fn execute_batch_or_rollback(
+        &self,
+        batch: Vec<Query>,
+    ) -> Result<(Vec<Option<QueryResult>>, State)> {
```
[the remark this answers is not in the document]
MarinPostma: there may or may not be an ongoing transaction. If there isn't then this is no_op, the rollback will fail and we ignore it, it there is an ongoing transaction, then we rollback

sqld/src/hrana/batch.rs
MarinPostma: this is done when the program is executed there: https://github.com/libsql/sqld/pull/307/files#diff-b7c90481e4da77d124d59b67dd5c2141bfba58deac7df43808defee65815fef9R308

sqld/src/database/write_proxy.rs
MarinPostma: That's right, I went a bit too fast over this. I pushed a commit that address this issue, lmk

sqld/src/database/mod.rs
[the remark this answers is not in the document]
honzasp: Yeah, I understand what the code is doing, but the reason for this behavior is not clear to me? (I'm not saying it's a problem, I would just like to understand :))

sqld/src/database/write_proxy.rs
```diff
@@ -115,50 +115,67 @@ impl WriteProxyDatabase {
...
-                    // Set state to invalid, so next call is sent to remote, and we have a chance
-                    // to recover state.
-                    *state = State::Invalid;
-                    Err(Error::RpcQueryExecutionError(e))
+        if *state == State::Init && pgm.is_read_only() {
+            // This program won't perform any write, we try to predict it's final state, i.e, if
+            // it's an interactive transaction or not, and then attempt to execute it. If it turns
+            // out that it was indeed an interactive transaction, we rollback the current
+            // transaction, and attempt to execute again on the remote
+            if predict_final_state(*state, pgm.steps.iter().map(|s| &s.query.stmt)) == State::Init {
```
honzasp, line 166: I think that this check will always fail for batches that are produced by libsql client. The batch looks like `BEGIN`, ..., `COMMIT`, `ROLLBACK`, so the `predict_final_state()` function will always return `State::Invalid`, right?

honzasp: If we verify that all statements are read-only, then perhaps we can always just execute the batch locally, and if this leaves a transaction open, then rollback it and retry on master? In effect, do the same as now, without checking for the predicted final state.

sqld/src/database/mod.rs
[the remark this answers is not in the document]
MarinPostma: It is up to the client to decide whether the content of the batch should be executed from within a transaction, the only concern of executing a batch from the HTTP perspective is that it doesn't leave a transaction open after all statements have been executed

MarinPostma:
bors merge

PR #323, opened 2023-04-07
Replica connection wait for replication

MarinPostma:
What this PR do is quite simple: it makes every replica database connection to wait for the replica instance to catch up with the last write that they performed.

When a replica detects a write, this write is proxied to the primary, that executes it, and then starts replicating the state. The replica receives a response that contains a logical timestamp issued after the write was performed. The replica's connection keep this timestamp and on subsequent operation on the _local_ connection, waits until the replicator has applied changes up to the stored timestamp.

Note that each connection holds a different timestamp, so a connection may proceed with a read, while another is still waiting.

The basic guarantee that it gives is read your write: a connection is always guaranteed to witness it's write. In fact the guarantee is much stronger because of the underlying consistency of sqlite, the real-time consistency model of sqld is now as strong as `sequential consistency`.

### DIFF
--- a/sqld/proto/proxy.proto
+++ b/sqld/proto/proxy.proto
@@ -89,6 +89,8 @@ message ExecuteResults {
     }
     /// State after executing the queries
     State state = 2;
+    /// Primary frame_no after executing the request.
+    uint64 current_frame_no = 3;
 }
 
 message Program {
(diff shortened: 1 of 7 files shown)


MarinPostma:
bors merge

PR #328, opened 2023-04-11
export sqlite3 dump

MarinPostma:
This PR ports the dumping loginc from sqlite to sqld. It allows the dumping of sqld tables to a sqlite compatible format.

sqld now exposes a CLI interface to export dumps, eg:

- print the dump for the database at the default path (`data.sqld`) to stdout:
```
sqld dump
```

- print the dump for the database at the path specified by `--db-path`:
```
sqld --db-path some/path/to/database dump
```

- write the dump for the database at the path specified by `--db-path` at the path specified by path:

```
sqld --db-path some/path/to/database dump --path write/dump/here
```

`dump` is a subcommand, so any argument passed after the `dump` have dump as subject, and any argument passed before `dump` have `sqld` as subject. Any argument with `sqld` as subject other than `--db-path` is ignored.

### DIFF
--- a/sqld/src/database/dump/exporter.rs
+++ b/sqld/src/database/dump/exporter.rs
@@ -120,7 +120,7 @@ impl<W: Write> DumpState<W> {
         let col_count = stmt.column_count();
         let mut rows = stmt.query(())?;
         while let Some(row) = rows.next()? {
-            let ValueRef::Text(sql) = row.get_ref(0)? else { bail!("first row in a table dump query should be of tyhpe text") };
+            let ValueRef::Text(sql) = row.get_ref(0)? else { bail!("the first row in a table dump query should be of type text") };
             self.writer.write_all(sql)?;
             for i in 1..col_count {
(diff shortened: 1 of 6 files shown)


MarinPostma:
@psarna requesting you for review on this one, since it's sqlite related

MarinPostma:
@athoscouto FYI

psarna (approved):
Looks good (I *really* hope I didn't miss anything in the port from C)

MarinPostma:
@psarna this is how we do it now, but it makes the images more complex... and we wanted to build other nice features on top of that, such as automatic migration etc, which makes if much more convenient. Another alternative is to move the dump code from shell.c to the libsql API?

MarinPostma:
> @psarna this is how we do it now, but it makes the images more complex... and we wanted to build other nice features on top of that, such as automatic migration etc, which makes if much more convenient. Another alternative is to move the dump code from shell.c to the libsql API?

MarinPostma:
bors merge

PR #331, opened 2023-04-14
whitelist permanent views

MarinPostma:
whitelist support for permanent views.

temporary view are still blacklisted, because they are tied to a specitif connection. It requires that we are able to sync the state of the connection on the primary and replica side,  which we are currently unable to do.


penberg (approved):

MarinPostma:
bors merge

PR #332, opened 2023-04-14
enable permanent triggers

MarinPostma:
enable permanent triggers. Temporary triggers are not allowed for the same reasons a #331


MarinPostma:
bors merge

PR #334, opened 2023-04-15
allow building from source tarball

glommer:
vergen by default will try to enable git, which means the build will fail if compiling from a source tarball.

Tools like homebrew have a preference towards building from the git tarball, so let's enable that.


glommer:
Marin, can we please tag a release once this is in ?

PR #335, opened 2023-04-16
feat: update pgwire to 0.13

sunng87:
This patch upgrade pgwire to 0.13. It includes some APIs change (mostly simplification)


sqld/src/postgres/handlers.rs
```diff
@@ -98,47 +94,85 @@ impl ExtendedQueryHandler for QueryHandler {
...
-
-        let query = Query { stmt, params };
-        let include_col_defs = client.metadata_mut().remove(REQUEST_DESCRIBE).is_some();
-        self.handle_queries(vec![query], include_col_defs)
-            .await
-            .map(|mut res| {
-                assert_eq!(res.len(), 1);
-                res.pop().unwrap()
-            })
+        unreachable!()
```
MarinPostma, line 169: I get that this is unreachable because we provide an implementation for `on_execute`?

sunng87: Yes. This is called by `on_execute`'s default implementation. However, we now use a custom implementation because of how we deal with response of `Describe` message. So `do_query` is no longer called.

MarinPostma:
Closing this since we discontinued support for Postgres wire. Thanks for your contributions @sunng87, much appreciated :)

PR #339, opened 2023-04-19
Enable extensions

glommer:
Extensions are loaded from a directory, and we require a list on that directory with the expected sha256.

The main disadvantage of this approach is that all replicas need to have the same list, and upon restarts, if we're adding extensions, this can get out of sync.

This is likely something that we can live with for now, and in the future implement a more robust mechanism where the list of approved extensions are listed on a table, for example.


sqld/src/database/libsql.rs
```diff
@@ -199,16 +202,29 @@ struct Connection {
 impl Connection {
     fn new(
         path: &Path,
+        extensions: Vec<PathBuf>,
         wal_hook: impl WalHook + Send + Clone + 'static,
         with_bottomless: bool,
         stats: Stats,
     ) -> anyhow::Result<Self> {
-        Ok(Self {
+        let x = Self {
```
honzasp, line 210: `x` is not a great variable name. Maybe `this` would be better? It is often used in Rust to refer to "self", when `self` cannot be used.

sqld/src/database/libsql.rs
```diff
@@ -199,16 +202,29 @@ struct Connection {
...
             stats,
-        })
+        };
+
+        for ext in extensions {
+            unsafe {
+                let _guard = rusqlite::LoadExtensionGuard::new(&x.conn).unwrap();
+                x.conn
+                    .load_extension(&ext, None)
+                    .with_context(|| format!("Loading extension: {}", &ext.display()))?;
```
honzasp, line 222: The anyhow context is printed as a standalone error message, so it will be less confusing if the error message says something like "Could not load extension: {}".

sqld/src/lib.rs
```diff
@@ -228,6 +237,53 @@ fn check_fresh_db(path: &Path) -> bool {
...
+        let mut list = File::open(&extensions_list).with_context(|| {
+            format!(
+                "While trying to read extensions: cannot load {}",
+                &extensions_list.display()
+            )
+        })?; // anyhow;
+        let mut file_contents = String::new();
+
+        list.read_to_string(&mut file_contents)
+            .with_context(|| format!("can't read {}", &extensions_list.display()))?;
```
honzasp, line 253: You can use `std::fs::read_to_string()` to open the file and read its content in a single step, something like `let file_contents = std::fs::read_to_string(&extensions_list).with_context(|| { ... })`

sqld/src/lib.rs
```diff
@@ -228,6 +237,53 @@ fn check_fresh_db(path: &Path) -> bool {
...
+
+        for line in extensions {
+            let ext_info: Vec<&str> = line.trim().split(' ').filter(|x| !x.is_empty()).collect();
+
+            let ext_sha = ext_info.first().ok_or_else(|| {
+                anyhow::anyhow!("invalid line on {}: {}", &extensions_list.display(), line)
+            })?;
+            let ext_fname = ext_info.get(1).ok_or_else(|| {
+                anyhow::anyhow!("invalid line on {}: {}", &extensions_list.display(), line)
+            })?;
```
honzasp, line 265: We can use `str.split_once()` here, something like `let (ext_sha, ext_fname) = line.trim().split_once(|c| c.is_whitespace()).with_context(...)`

honzasp (commented):
Looks OK, just a few nitpicks. However, why do we need to check the sha256 checksum of the extensions? Is there some scenario where `sqld` and the `trusted.lst` are distributed separately from the extensions, so we need to verify that the extensions are authentic?

glommer:
> Is there some scenario where `sqld` and the `trusted.lst` are distributed separately from the extensions, so we need to verify that the extensions are authentic?

I am being extra careful because most sqlite extensions are plain C and a security risk in case you end up loading the wrong thing. It has a nice side effect of validating before the connection is open (after the connection is open it is too late to throw good errors) that the extensions are indeed present and can be loaded.

glommer:
@MarinPostma @honzasp I force-pushed this

sqld/src/lib.rs
MarinPostma: split_once and other method act on bytes and not char: that be surprising sometimes.

sqld/src/lib.rs
```diff
@@ -228,6 +234,51 @@ fn check_fresh_db(path: &Path) -> bool {
...
 
+fn validate_extensions(extensions_path: Option<PathBuf>) -> anyhow::Result<Vec<PathBuf>> {
+    let mut valid_extensions = vec![];
+    if let Some(ext_dir) = extensions_path {
+        let extensions_list = ext_dir.join("trusted.lst");
+
+        let file_contents = std::fs::read_to_string(&extensions_list)
+            .with_context(|| format!("can't read {}", &extensions_list.display()))?;
+
+        let extensions = file_contents.split_terminator('\n');
```
MarinPostma, line 245: nit: there's a `lines` methods on `str`: https://doc.rust-lang.org/std/primitive.str.html#method.lines

glommer: Thanks. I thought that split_terminator would work better because it doesn't include the final newline in the file so I didn't have to filter. But it turns out it doesn't coalesce them, so I actually have to filter anyway (which I am not doing). Will update, using lines + filter.

sqld/src/lib.rs
honzasp: Ok, I thought `|c| c.is_whitespace()` would match a stretch of whitespace characters, but you are right, it matches just a single character. However, it does work on characters, not on bytes (it handles UTF-8 correctly; after all, this is Rust! :))

honzasp:
Looks good to me, but I will let Marin make the final verdict, I'm not sure that I understand all implications that extensions have on replication and stuff :)

MarinPostma:
bors merge

PR #341, opened 2023-04-20
Enable Hrana if only HTTP is active

honzasp:
Previously, we needed to enable a dedicated Hrana port to enable WebSocket upgrades over HTTP. With this change, enabling HTTP is enough to enable the upgrades.


MarinPostma (approved):
bors merge

PR #347, opened 2023-04-24
disable temp table creation

MarinPostma:
Temp tables cannot be supported for now because they change the state of the connection.


MarinPostma:
bors merge

MarinPostma:
bors merge

PR #384, opened 2023-05-08
limit max response size

MarinPostma:
This PR adds a limit to the size of the response that sqld can send. This is settable with the `--max-response-size` arg, and defaults to 10MB. If the size for a response would exceed this number, an error is returned to the user. The size of the payload is estimated before the payload is serialized, so it's a best effort approach. We might over or underestimate the actually payload size.

### DIFF
--- a/sqld/src/error.rs
+++ b/sqld/src/error.rs
@@ -29,6 +29,8 @@ pub enum Error {
     ReplicatorExited,
     #[error("Timed out while openning database connection")]
     DbCreateTimeout,
+    #[error("Response payload exceeds the maximum allowed size, consider limiting the retreived results.")]
+    ResponseTooLarge,
 }
 
 impl From<tokio::sync::oneshot::error::RecvError> for Error {
--- a/sqld/src/query.rs
+++ b/sqld/src/query.rs
@@ -120,6 +120,12 @@ pub struct Row {
     pub values: Vec<Value>,
 }
 
+impl Row {
+    pub fn size_hint(&self) -> usize {
+        self.values.iter().map(|v| v.size_hint()).sum()
+    }
+}
+
 /// Mirrors rusqlite::Value, but implement extra traits
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum Value {
@@ -128,6 +134,21 @@ pub enum Value {
     Real(f64),
     Text(String),
     Blob(Vec<u8>),
+}
+
+impl Value {
+    /// Approx. of the size in bytes required to store the value as json
+    fn size_hint(&self) -> usize {
+        match self {
+            Value::Null => 4,
+            // ~ number of digits needed to encode i
+            Value::Integer(i) => i.ilog10() as _,
+            // find better estimation: size of ascii u64::MAX
+            Value::Real(_) => "18446744073709551615".len(),
+            Value::Text(s) => s.len(),
+            Value::Blob(b) => b.len(),
+        }
+    }
 }
 
 impl From<rusqlite::types::Value> for Value {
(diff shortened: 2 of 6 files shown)


athoscouto:
@psarna @honzasp any of you feel comfortable reviewing this?
I want to move forward and merge it since I'm missing this functionality.

sqld/src/error.rs
```diff
@@ -29,6 +29,8 @@ pub enum Error {
     ReplicatorExited,
     #[error("Timed out while openning database connection")]
     DbCreateTimeout,
+    #[error("Response payload exceeds the maximum allowed size, consider limiting the retreived results.")]
```
CodingDoug, line 32: Some things about this message:

1. Since we're not calculating the size exactly, we should be more clear about that.  It's actually the *estimated* size that we're limiting, based on the number of rows (and the estimated size of each row?).
2. The maximum size itself isn't stated. Could we give some indicator here?
3. It would be great if this message could be formatted to indicate the number of actual rows that would theoretically be allowed to pass for the query that failed. It looks like we know that from `execute_query_inner`?  Without some information here, the reader of this message doesn't really know where to start truncating, and would have to experimentally bisect until the query works.

sqld/src/query.rs
```diff
@@ -130,6 +136,21 @@ pub enum Value {
...
+    /// Approx. of the size in bytes required to store the value as json
+    fn size_hint(&self) -> usize {
+        match self {
+            Value::Null => 4,
+            // ~ number of digits needed to encode i
+            Value::Integer(i) => i.ilog10() as _,
+            // find better estimation: size of ascii u64::MAX
+            Value::Real(_) => "18446744073709551615".len(),
+            Value::Text(s) => s.len(),
+            Value::Blob(b) => b.len(),
```
honzasp, line 149: We base64-encode blobs, so this should be `(b.len()+2)/3*4`

sqld/src/query.rs
```diff
@@ -130,6 +136,21 @@ pub enum Value {
     Blob(Vec<u8>),
 }
 
+impl Value {
+    /// Approx. of the size in bytes required to store the value as json
+    fn size_hint(&self) -> usize {
```
honzasp, line 141: This function only estimates size of the value itself, but the JSON representation also includes the type (i.e., integers are represented as `{"type":"integer","value":"123"}`. I think that we can improve the estimate by adding 30 to the value returned from the `match`.

sqld/src/query.rs
```diff
@@ -120,6 +120,12 @@ pub struct Row {
     pub values: Vec<Value>,
 }
 
+impl Row {
+    pub fn size_hint(&self) -> usize {
+        self.values.iter().map(|v| v.size_hint()).sum()
```
honzasp, line 125: When we are at it, we can also estimate the overhead of row encoding, which we can simplify to be 1 additional byte per column (for the comma that separates values in an array) and 3 additional bytes per row (for the `[]` that delimit the array of columns and for the comma that separates rows).

This overhead might be meaningful if the query returns a huge number of rows.

honzasp (changes requested):
Looks ok, but the size of JSON values is now underestimated a lot, because we don't account for the type information (i.e., an integer 123 is encoded as `{"type":"integer","value":"123"}`, not as `123`).

sqld/src/query.rs
```diff
@@ -130,6 +136,21 @@ pub enum Value {
     Blob(Vec<u8>),
 }
 
+impl Value {
+    /// Approx. of the size in bytes required to store the value as json
+    fn size_hint(&self) -> usize {
+        match self {
+            Value::Null => 4,
```
honzasp, line 143: Once we account for the type information, the additional cost of NULL is zero (NULL is encoded as `{"type":"null"}`).

MarinPostma: uhmm that's only true for hrana. Let me think of a better solution

honzasp: Do we still need to care about the legacy HTTP API? It even makes sense that both APIs apply the response size limit in the same way.

MarinPostma:
closing in favor of #451

PR #385, opened 2023-05-08
Additions to Hrana 2

honzasp:
Profusion's new Python client has highlighted the need to add two more features to Hrana 2:

- Include decltype of every column in statement response.
- Add a `sequence` request to execute a sequence of SQL statements separated by semicolons.

### DIFF
--- a/docs/HRANA_2_SPEC.md
+++ b/docs/HRANA_2_SPEC.md
@@ -33,17 +33,19 @@ connection properly authenticated.
 
 ## Requests
 
-Version 2 introduces three new requests:
+Version 2 introduces four new requests:
 
 ```typescript
 type Request =
     | ...
+    | SequenceReq
     | DescribeReq
     | StoreSqlReq
     | CloseSqlReq
 
 type Response =
     | ...
+    | SequenceResp
     | DescribeResp
     | StoreSqlReq
     | CloseSqlReq
@@ -64,7 +66,7 @@ type StoreSqlResp = {
 ```
 
 The `store_sql` request stores an SQL text on the server. The client can then
-refer to this SQL text in `Stmt` objects by its id, instead of repeatedly
+refer to this SQL text in other requests by its id, instead of repeatedly
 sending the same string over the network.
 
 SQL text ids are arbitrary 32-bit signed integers assigned by the client. It is
@@ -90,6 +92,29 @@ the response.
 
 It is not an error if the client attempts to close a SQL text id that is not
 used.
+
+### Execute a sequence of SQL statements
+
+```typescript
+type SequenceReq = {
+    "type": "sequence",
+    "stream_id": int32,
+    "sql"?: string | null,
+    "sql_id"?: int32 | null,
+}
+
+type SequenceResp = {
+    "type": "sequence",
+}
+```
+
+The `sequence` request executes a sequence of SQL statements separated by
+semicolons on the stream given by `stream_id`. `sql` or `sql_id` specify the SQL
+text; exactly one of these fields must be specified.
+
+Any rows returned by the statements are ignored. If any statement fails, the
+subsequent statements are not executed and the request returns an error
+response.
 
 ### Describe a statement
 
@@ -180,3 +205,15 @@ id that has previously been stored with the `store_sql` request. Exactly one of
 `sql` and `sql_id` must be passed.
 
 Also, the `want_rows` field is now optional and defaults to `true`.
+
+### Statement result
+
+```typescript
+type Col = {
+    "name": string | null,
+    "decltype": string | null,
+}
+```
+
+In version 2 of the protocol, the column descriptor in the statement result also
+includes the declared type of the column (if available).
--- a/sqld/src/hrana/proto.rs
+++ b/sqld/src/hrana/proto.rs
@@ -29,6 +29,7 @@ pub enum Request {
     CloseStream(CloseStreamReq),
     Execute(ExecuteReq),
     Batch(BatchReq),
+    Sequence(SequenceReq),
     Describe(DescribeReq),
     StoreSql(StoreSqlReq),
     CloseSql(CloseSqlReq),
@@ -41,6 +42,7 @@ pub enum Response {
     CloseStream(CloseStreamResp),
     Execute(ExecuteResp),
     Batch(BatchResp),
+    Sequence(SequenceResp),
     Describe(DescribeResp),
     StoreSql(StoreSqlResp),
     CloseSql(CloseSqlResp),
@@ -84,6 +86,17 @@ pub struct BatchResp {
     pub result: BatchResult,
 }
 
+#[derive(Deserialize, Debug)]
+pub struct SequenceReq {
+    pub stream_id: i32,
+    #[serde(default)]
+    pub sql: Option<String>,
+    #[serde(default)]
+    pub sql_id: Option<i32>,
+}
+
+#[derive(Serialize, Debug)]
+pub struct SequenceResp {}
 #[derive(Deserialize, Debug)]
 pub struct DescribeReq {
     pub stream_id: i32,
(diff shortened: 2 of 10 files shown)


docs/HRANA_2_SPEC.md
```diff
@@ -91,6 +93,29 @@ the response.
...
+type SequenceResp = {
+    "type": "sequence",
+}
+```
+
+The `sequence` request executes a sequence of SQL statements separated by
+semicolons on the stream given by `stream_id`. `sql` or `sql_id` specify the SQL
+text; exactly one of these fields must be specified.
+
+Any rows returned by the statements are ignored. If any statement fails, the
```
haaawk, line 115: Why are we not returning any rows? if not for all select statements then at least for the last one?

sqld/src/hrana/proto.rs
```diff
@@ -84,6 +86,17 @@ pub struct BatchResp {
     pub result: BatchResult,
 }
 
+#[derive(Deserialize, Debug)]
+pub struct SequenceReq {
+    pub stream_id: i32,
+    #[serde(default)]
+    pub sql: Option<String>,
+    #[serde(default)]
+    pub sql_id: Option<i32>,
```
haaawk, line 95: Is there a way to make serde deserialize sql and sql_id fields into an Enum?

docs/HRANA_2_SPEC.md
honzasp: The `sequence` request is designed to implement client APIs like:
- [`Cursor.executescript()`](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executescript) in Python
- [`Database.exec()`](https://github.com/TryGhost/node-sqlite3/wiki/API#execsql--callback) in Node.js `sqlite3`
- [`Connection.execute_batch()`](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.execute_batch) in Rust `rusqlite`

These methods don't return any rows, and some of them even report an error if some of the statements do! Therefore, to keep the protocol and the implementation simple, no results are returned in Hrana 2. We can always extend the protocol to optionally return the results later, if it turns out that we need it.

sqld/src/hrana/proto.rs
honzasp: It might be possible to do this by using [`#[serde(flatten)]`](https://serde.rs/field-attrs.html#flatten) and some manual serializer/deserializer implementation, but I decided to keep the structures simple. There is a helper method that extracts the SQL text from `sql` xor `sql_id` or produces an error if not exactly one field was given.

haaawk (approved):

docs/HRANA_2_SPEC.md
haaawk: execute_batch could return results. I'm asking because turso shell will support executing batch of statements separated with ; and it is useful to return values. For example BEGIN; INSERT INTO t ...; SELECT FROM t; COMMIT

honzasp: Won't the shell need to parse the semicolons anyway, to detect when the input is finished?

honzasp:
> Could we support parameters in sequence?

Not in any meaningful way, because parameters are applied to individual statements. What would be the meaning of `SELECT ?; SELECT ?` and `SELECT :a; SELECT :a`? That's the main reason why no API provides this feature.

docs/HRANA_2_SPEC.md
haaawk: it may parse it but it will have to send all statements as a single batch in the end anyway

haaawk:
> > Could we support parameters in sequence?
> 
> Not in any meaningful way, because parameters are applied to individual statements. What would be the meaning of `SELECT ?; SELECT ?` and `SELECT :a; SELECT :a`? That's the main reason why no API provides this feature.

Yeah. I know this is tricky but there are also totally valid cases and I'm not sure it is ok to deny all just because there are cases we don't know how to handle.

I think mostly of a case where we can't have interactive transactions and sequence/batch is the only way to do transactions. In such cases people will be looking for parameters for their queries.

honzasp:
> Yeah. I know this is tricky but there are also totally valid cases and I'm not sure it is ok to deny all just because there are cases we don't know how to handle.
> 
> I think mostly of a case where we can't have interactive transactions and sequence/batch is the only way to do transactions. In such cases people will be looking for parameters for their queries.

Neither SQLite nor Postgres support this, so our users can probably live without this ill-defined feature as well. If you need to programmatically issue multiple statements, you should be using `batch` (which supports parameters), not `sequence`.

docs/HRANA_2_SPEC.md
honzasp: To send a batch of statements, the shell should use the `batch` request.

barbieri: I like @haaawk's idea to allow for `want_rows=true` in the request and receive something similar to `BatchResult`, but we'd need another array with the statements, such as it would be implemented as a `batch` where every line depends on the previous not failing.

```typescript
type SequenceResult = BatchResult & { "stmts": string[] };
```

haaawk:
> > Yeah. I know this is tricky but there are also totally valid cases and I'm not sure it is ok to deny all just because there are cases we don't know how to handle.
> > I think mostly of a case where we can't have interactive transactions and sequence/batch is the only way to do transactions. In such cases people will be looking for parameters for their queries.
> 
> Neither SQLite nor Postgres support this, so our users can probably live without this ill-defined feature as well. If you need to programmatically issue multiple statements, you should be using `batch` (which supports parameters), not `sequence`.

Why can batch support parameters and sequence not?

docs/HRANA_2_SPEC.md
haaawk: batch is not providing the right semantic. `BEGIN;INSERT...;SELECT...;COMMIT` run in batch would not be atomic would it?
I could also want `BEGIN;...;COMMIT;BEGIN;...;COMMIT`. It would definitely not work as expected with batch, will it @honzasp ?

honzasp:
> Why can batch support parameters and sequence not?

Because in the `batch` request, the statements are passed as separate units of SQL text with parameters. But in the `sequence` request, the payload is an opaque blob of SQL, and there is no clear way to match arguments in the request to parameters in the SQL statements.

If the statements are generated programmatically and may thus contain parameters, `sequence` is not the right tool, `batch` should be used instead. `sequence` is intended to be used to execute manually written SQL scripts, SQL dumps and the like.

docs/HRANA_2_SPEC.md
honzasp: What would be the use for this?

haaawk:
I don't think we should merge this until we agree on the API. Current API seems like it will have to change to support the use cases we need so I don't think it's a good idea to make it available for users.

barbieri:
> > Why can batch support parameters and sequence not?
> 
> Because in the `batch` request, the statements are passed as separate units of SQL text with parameters. But in the `sequence` request, the payload is an opaque blob of SQL, and there is no clear way to match arguments in the request to parameters in the SQL statements.
> 
> If the statements are generated programmatically and may thus contain parameters, `sequence` is not the right tool, `batch` should be used instead. `sequence` is intended to be used to execute manually written SQL scripts, SQL dumps and the like.

can't you simply assume the same parameters for all the statements? Positional would suck to use, but named would be ok:

```typescript
type SequenceReq = {
    "type": "sequence",
    "stream_id": int32,
    "sql"?: string | null,
    "sql_id"?: int32 | null,
    "args"?: Array<Value>, // wouldn't hurt, but not as usable
    "named_args"?: Array<NamedArg>, // works all right
    "want_rows": boolean,
}
```

```sql
SELECT bla WHERE id = @id;
UPDATE SET bla = @new WHERE id = @id;
````

Ok, can be done with a `batch`, but I'd see `sequence` as a shortcut to generate a `batch`, implementing on top of the same code server-side... just relieves the parsing of statements from client side.

PR #388, opened 2023-05-10
Minor dockerfile and readme improvements

nukemberg:
Minor improvements to dockerfile and README:
- Docker users expect volumes for data files, especially for databases
- Although containers are "isolated" it's still a good practice to run as a non-root user
- Build instructions using docker, probably easier for beginners, CI, etc.


penberg:
Thanks for the contribution, @nukemberg!

PR #392, opened 2023-05-12
Add `libsql` namespace to docker run command

avinassh:
Added `libsql` namespace to docker run command. Otherwise it breaks with following error:

```shell
$ docker container run --rm -it sqld /bin/sqld --help

Unable to find image 'sqld:latest' locally

docker: Error response from daemon: pull access denied for sqld, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```


MarinPostma:
thanks!

MarinPostma (approved):

PR #403, opened 2023-05-17
Stateful Hrana-over-HTTP API

honzasp:
WebSockets are not reliably implemented in all environments that we want to support (such as Miniflare or Vercel Edge Functions), so we need an HTTP fallback for Hrana. Unfortunately, our existing HTTP API provides a limited functionality compared to Hrana; most importantly, it does not support interactive transactions.

This PR proposes a new HTTP API that is more worthy of the name "Hrana over HTTP": it supports stateful streams and implements the same requests as Hrana. It is expected that Hrana client libraries will support both WebSockets and HTTP with similar interfaces.

HTTP is fundamentally stateless, so using it as the transport layer for a stateful protocol (Hrana streams) brings two challenges:

1. **Request ordering**: HTTP does not guarantee any ordering between concurrent requests. This PR solves it by forcing clients to issue requests serially. However, we support rudimentary pipelining: one HTTP request can contain a sequence of Hrana requests.
2. **Sticky sessions**: with HTTP, it's not uncommon that requests to the same host reach different server processes, due to load balancing, geographical routing or other complications of modern infrastructure. This PR works around this issue by using @barbieri-s idea: when the client opens a stream, the server may return an URL that the client will use in subsequent requests. This URL will specify the particular server instance that has received this request.

(This PR replaces the recently introduced HTTP API v2. We are not yet using this API in any clients, so we can remove it without breaking anything.)


docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
 -> {
-    "stmt": Stmt,
+    "baton": string | null,
+    "requests": Array<StreamRequest>,
 }
 
 <- {
-    "result": StmtResult,
+    "baton": string | null,
+    "base_url": string | null,
```
athoscouto, line 49: Have you considered answering the first POST with `201` and a `Location` header?
Not sure that is appropriate for this use case though.

docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
+response for a GET request on `/v2`. This can be used as a crude version
+negotiation mechanism by the client.
 
-## Execute a statement
+## Execute requests on a stream
 
-```
-POST /v2/execute
+```typescript
+POST /v2/stream
```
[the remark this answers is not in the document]
honzasp: The reason for this name is that this sends requests to a Hrana stream. "execute" means "execute a single statement" in Hrana, so that might be confusing, too :/

docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
-All request and response bodies are encoded in JSON, with content type
-`application/json`.
+Individual requests on the same stream are tied together by the use of a baton.
+The server returns a baton in every response to a request on the stream, and the
+client then needs to include the baton in the subsequent request. The client
+must serialize the requests: it must wait for a response to the previous request
+before sending next request.
 
-## Version check
+The server can also optionally specify a different URL that the client should
```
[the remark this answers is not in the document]
honzasp: The `base_url` is optional, if the server returns `null` then it has no effect. This means that if we find another solution, we can just stop using `base_url` without any changes to the protocol or to the clients.

The baton is an opaque string for the purposes of this protocol, so we can encode an identifier of the sqld process into it and use it to redirect a mis-routed HTTP request. Do you think that we should just use this approach and drop the the `base_url` mechanism?

Btw, do you have some example of an HTTP client that gives control over the underlying TCP connection? All HTTP clients that we use in client libraries (JavaScript's `fetch`, Python's `asyncio` or `requests`, Rust's `hyper`/`reqwest`, Golang's `net/http`) don't seem to provide this level of control.

docs/HTTP_V2_SPEC.md
honzasp: Redirecting the client would add another network roundtrip for the first request.

docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
+request specifies the stream. If the client sets `baton` to `null`, the server
+should create a new stream.
+
+Server responds with another `baton` value in the response. If the `baton` value
+in the response is `null`, it means that the server has closed the stream. The
+client must use this value to refer to this stream in the next request (the
+`baton` in the response should be different from the `baton` in the request).
+This forces the client to issue the requests serially: it must wait for the
+response from a previous `stream` request before issuing another request on the
+same stream.
```
barbieri, line 78: as per my above example, I don't see why we need to bother with baton if the client side can trust serialization on its side, either by waiting on previous requests or natively if the language doesn't support any parallel execution.

just makes the client implementation harder.

docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
+
+The server should ensure that the `baton` values are unpredictable and
+unforgeable, for example by cryptographically signing them.
+
+If the `base_url` in the response is not `null`, the client should use this URL
+when sending further requests on this stream. If it is `null`, the client should
+use the same URL that it has used for the previous request. The `base_url`
+must be an absolute URL with "http" or "https" scheme.
```
barbieri, line 86: as per above, you can force the url and avoid base, no composition needs to be done, just get the opaque value and use it as the url. No headers or url assembling is needed.

you can use it in shell scripts and barebone clients with ease

docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
+before sending next request.
 
-## Version check
+The server can also optionally specify a different URL that the client should
+use for the requests on the stream. This can be used to ensure that stream
+requests are "sticky" and reach the same server.
 
-```
+The server will close streams after a short period of inactivity, to make sure
+that abandoned streams don't accumulate on the server.
```
barbieri, line 25: since we should be closing the stream if the client closes it (TCP), I'd highly recommend we allow the client to specify the `timeout=seconds` in the url param. Of course, default to some reasonable timeout, but let users do long-lived sessions whenever needed.

In addition to that, if we want to be resilient to quick network failures, we could allow the `session` in the control-url returned in my example below to be given as another url param, ex: `reconnect=session`

Then connections that are not explicitly closed via control_url (ie: TCP disconnect/timeout detected by the server) will be put in a "disposal queue" instead of closed immediately.

If the new stream is created with the `reconnect=session` as parameter, then it will see if it's still pending disposal and will bring it to use again.

New control requests should also remove from the disposal queue and bring it to life again instead of failing (suppose the stream fails tcp/networking, but a controls comes in, then the stream reconnects).

Whenever the reconnection happens, the pending responses are flushed to the stream.

This reconnection timeout should also be configurable, or at least enabled/disabled... in some scenarios this may be undesired.

NOTE: to reliably check the client, the server should be sending some `: ping\n\n` messages to the client, they should be ignored (https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format), but this should lead the tcp to fail if the client is gone.

docs/HTTP_V2_SPEC.md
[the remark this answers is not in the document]
honzasp: With this proposed API, no server-sent events are needed, these are standard HTTP requests, for maximum simplicity and compatibility.

docs/HTTP_V2_SPEC.md
honzasp: Why introduce another roundtrip if we don't need it? Low latency is what Turso is all about, and network roundtrips are the dominating factor in latency.

docs/HTTP_V2_SPEC.md
honzasp: The baton is used to identify the stream, while the `base_url` is just an optional mechanism to ensure that sessions are sticky. As Athos pointed out, we may use a different approach in the future, so it makes sense to implement these orthogonal features with orthogonal mechanisms.

docs/HTTP_V2_SPEC.md
honzasp: The client needs to do URL resolution anyway: the user passes an absolute URL like "https://my.turso.io" and the client must add "v2/stream" to obtain "https://my.turso.io/v2/stream". When the server returns a `base_url`, the client just replaces the user-provided URL with the `base_url`.

honzasp: (and there are no headers involved, everything is passed in request and response bodies as JSON)

docs/HTTP_V2_SPEC.md
```diff
@@ -1,106 +1,236 @@
...
+
+The `requests` array in the request specifies a sequence of stream requests that
+should be executed on the stream. The server executes them in order and returns
+the results in the `results` array in the response. Result is either a success
+(`type` set to `"ok"`) or an error (`type` set to `"error"`). The server always
+executes all requests, even if some of them return errors.
+
+If the client receives an HTTP error (4xx or 5xx response) in response to the
+`stream` endpoint, it means that the server encountered an internal error and
+the stream is no longer valid.
```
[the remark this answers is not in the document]
honzasp: Yes, usually, but the server can fail with unexpected errors, too. (A failure to open the database is also an unexpected error, btw :))

docs/HTTP_V2_SPEC.md
honzasp: The only point of keeping a long-living TCP connection would be to close the stream immediately when the connection is broken. Introducing the disposal queue would nullify this advantage.

Also, environments that can establish a reliable long-living TCP connection can just use WebSockets. The HTTP API is designed for environments where we can only perform simple HTTP requests.

docs/HTTP_V2_SPEC.md
barbieri: what round trip you're taking about. In my proposal there is no extra roundtrips.

If you account about the initial redirect, you can understand it as an addressing resolution. You can avoid it if you want to add more logic to the load balancer/reverse proxy, then you keep the state in there and always return the same URL, without any redirect.

docs/HTTP_V2_SPEC.md
barbieri: and what's the advantage of base_url over a fully ready control url, with it all. And it's not optional, so it's consistent, can (should be) the same.

honzasp: The advantage is that the server does not have to know its URL.

docs/HTTP_V2_SPEC.md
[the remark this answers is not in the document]
honzasp: Yes, that's true. The server only produces 4xx error when the client violates the protocol (for example by sending invalid JSON) and 5xx error when it encounters an internal error. "Expected errors" (such as invalid SQL string or transaction timeout) are reported as errors inside Hrana.

docs/HTTP_V2_SPEC.md
honzasp: I don't understand the first sentence :)

Long polling is more complex and more fragile than "normal" HTTP requests and we can implement everything that we need without long polling, so why use it?

docs/HTTP_V2_SPEC.md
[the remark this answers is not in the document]
honzasp: I renamed this endpoint to "pipeline", because it receives a pipeline of requests on a stream. Does that sound better to you, Athos? @athoscouto

sqld/src/hrana/http/stream.rs
```diff
@@ -0,0 +1,402 @@
...
+            let mut state = server.stream_state.lock();
+            let stream = Box::new(Stream {
+                db: Some(db),
+                sqls: HashMap::new(),
+                stream_id: gen_stream_id(&mut state),
+                // initializing the sequence number randomly makes it much harder to exploit
+                // collisions in batons
+                baton_seq: rand::random(),
+            });
+            state.handles.insert(stream.stream_id, Handle::Acquired);
```
MarinPostma, line 161: If I understand correctly, every request, whether it explicitly asked for it or not, gets a stream allocated. That stream will be released as, and only expire in 10s. I think this is quite problematic.

MarinPostma (changes requested):
There is a fundamental flaw in this approach. No matter how we go about it, we need to limit the number of concurrently running connections. This protocol creates a stream database connection by default and holds onto it for a non-negligible period of time (10 s currently).
This leaves the duty to close the stream to the client, which, if he does, cost an extra roundtrip, and if he doesn't creates contention on the pool of connections.

I expect that most of the time, people will use single queries, and not close the stream.

As I expected, running some tests with high-concurrency, the first few requests (up until we reach the concurrency limit) get served immediately, all of the remaining request receive a timeout for failing to allocate a connection.

sqld/src/hrana/http/stream.rs
```diff
@@ -0,0 +1,402 @@
...
+                    if stream.baton_seq != baton_seq {
+                        return Err(ProtocolError::BatonReused).context(format!(
+                            "Expected baton seq {}, received {baton_seq}",
+                            stream.baton_seq
+                        ));
+                    }
+                }
+            };
+
+            let Handle::Available(mut stream) = mem::replace(handle.unwrap(), Handle::Acquired) else {
```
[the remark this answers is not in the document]
honzasp: Yeah, I would of course do this if I could, but how? In the `Some(Handle::Available(stream))` case of the `match`, is there a way to return the `stream` (which is `&mut Box<Stream>`) and replace the handle with `Handle::Acquired`?

sqld/src/hrana/http/request.rs
```diff
@@ -0,0 +1,111 @@
...
+                .map_err(catch_stmt_error)?;
+            proto::StreamResponse::Describe(proto::DescribeStreamResp { result })
+        }
+        proto::StreamRequest::StoreSql(req) => {
+            let sqls = stream_guard.sqls_mut();
+            let sql_id = req.sql_id;
+            if sqls.contains_key(&sql_id) {
+                bail!(ProtocolError::SqlExists { sql_id })
+            }
+            sqls.insert(sql_id, req.sql);
```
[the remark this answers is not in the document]
honzasp: Yeah, we should do this for WebSockets, too. Will fix.

sqld/src/hrana/http/stream.rs
honzasp: Yes, you cannot execute SQL statements without an SQL connection (= stream). If the client doesn't want to keep the stream around (for example, in `Client.execute()` and `Client.batch()`), then it can immediately close it adding the `close` request in the pipeline.

honzasp:
> There is a fundamental flaw in this approach.

Let me make this clear: I don't love this approach, I'm introducing this API only because Glauber strongly wants it, and he wants it because Miniflare has sketchy support for WebSockets and Vercel Edge Functions have no support at all, so we need an alternative.

I'm not arguing that this is a _great_ API, but I still think it's the best SQL connection API that we can have over HTTP.

> This leaves the duty to close the stream to the client, which, if he does, cost an extra roundtrip

That's not accurate: the close request can be pipelined as any other request. In particular, this means that in libsql client, `execute()` and `batch()` still need just a single roundtrip, because they close the stream immediately after it is created, and the `transaction()` method will pipeline the close request after the COMMIT/ROLLBACK statement. In both cases, we have the same number of roundtrips as with WebSockets.

> No matter how we go about it, we need to limit the number of concurrently running connections. This protocol creates a stream database connection by default and holds onto it for a non-negligible period of time (10 s currently). This leaves the duty to close the stream to the client, which, if he does, cost an extra roundtrip, and if he doesn't creates contention on the pool of connections.

Yes, but this is exactly the same situation as with WebSockets: the client is also responsible for closing the streams, and with WebSockets there is not even a timeout, so nothing prevents the clients from keeping the stream open forever. The only difference is that when the server learns that a WebSocket connection is broken, all its streams are closed immediately.

> I expect that most of the time, people will use single queries, and not close the stream.

People won't be sending these HTTP requests by hand, but they will use a libsql client, which will close the stream for them. The only way to leak streams with the libsql client is to create a `transaction()` and leak it without committing, rolling back or closing.

sqld/src/hrana/http/stream.rs
```diff
@@ -0,0 +1,402 @@
...
+                    return Err(StreamError::StreamExpired)
+                        .context(format!("Stream handle for {stream_id} is expired"));
+                }
+                Some(Handle::Available(stream)) => {
+                    if stream.baton_seq != baton_seq {
+                        return Err(ProtocolError::BatonReused).context(format!(
+                            "Expected baton seq {}, received {baton_seq}",
+                            stream.baton_seq
+                        ));
+                    }
```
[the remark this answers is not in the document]
honzasp: Not closing the stream here allows to client to safely retry HTTP requests: if you send the same request three times, the baton_seq check ensures that the request is executed at most once, and returning an error without closing the stream ensures that the client won't inadvertedly close the stream by retrying.

If a client does not pass the correct baton by mistake, it is a very obvious bug, so don't think that we should be too worried about encountering such clients on production.

sqld/src/hrana/http/stream.rs
```diff
@@ -0,0 +1,402 @@
...
+    let mut baton_data = [0; 48];
+    baton_data[0..16].copy_from_slice(&payload);
+    baton_data[16..48].copy_from_slice(&mac);
+    BASE64_STANDARD_NO_PAD.encode(baton_data)
+}
+
+/// Decodes a baton encoded with `encode_baton()` and returns `(stream_id, baton_seq)`. Always
+/// returns a [`ProtocolError::BatonInvalid`] if the baton is invalid, but it attaches an anyhow
+/// context that describes the precise cause.
+fn decode_baton(server: &Server, baton_str: &str) -> Result<(u64, u64)> {
```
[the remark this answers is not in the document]
honzasp: These are plain functions, what makes them look like methods that pretend to be methods? In general, I prefer to use plain function, unless there is a good reason to use a method.

For example, with the `Guard` struct it is very convenient to call functions like `let db = stream_guard.get_db()?` (rather than `let db = stream::guard_get_db(stream_guard)?`), but with functions like `gen_stream_id()`, `decode_baton()` and `pump_expire()`, I see no need to write them as `impl`-s on their first argument; it would only add syntactical noise (the `impl` and the extra indentation) and make it a teeny bit harder to follow.

In fact, that's one thing that I really like about Rust: it allows me to write code as modules with functions (and use methods when it's convenient), but it does not force me to implement everything as a method, like Java.

docs/HTTP_V2_SPEC.md
barbieri: > I don't understand the first sentence :)

the environment (aws lambda, azure functions and whatever runs on top of them, like vercel) cannot detect if the HTTP request is meant to be used in a long polling or not. All they can (and do) is impose some timeout on the request.


> Long polling is more complex and more fragile than "normal" HTTP requests and we can implement everything that we need without long polling, so why use it?

because you can detect failures and misbehavior. I also don't think it's complex, unless we're imagining different things

docs/HTTP_V2_SPEC.md
barbieri: But that's the easiest part. While doing that, it also helps the reverse proxy to properly connect.

barbieri:
> There is a fundamental flaw in this approach. No matter how we go about it, we need to limit the number of concurrently running connections. This protocol creates a stream database connection by default and holds onto it for a non-negligible period of time (10 s currently). This leaves the duty to close the stream to the client, which, if he does, cost an extra roundtrip, and if he doesn't creates contention on the pool of connections.

This is why I proposed the first connection is kept open as such handle, when it's gone, it's all closed... it can be due a network error, power failure, etc. See https://github.com/libsql/sqld/pull/403#discussion_r1197128267

I also mention that the timeout should be configurable otherwise you can't do longer tasks such as large dumps/loads. The default can be set to a reasonably low value such as 5s, but clients can ask for more when needed. Of course server can cap that.

> I expect that most of the time, people will use single queries, and not close the stream.

For those I think the batch/execute (existing) API can be kept as is, not using the pipeline/stream. Problem is how a generic client code will know that, say when integrating with ORM.

MarinPostma (approved):

MarinPostma (approved):

PR #434, opened 2023-05-25
Release pipeline

dyasny:
Create a gh action that builds releases on push to a proper tag

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,8 @@ on:
         description: 'Should this be a pre-release?'
         default: 'false'
         required: false
-
+  push:
+    branch: 'release-pipeline'
 jobs:
   build:
     # Build on Linux, Windows and MacOS
@@ -30,7 +31,8 @@ jobs:
 
       - uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.inputs.tag }}
+          #ref: ${{ github.event.inputs.tag }}
+          ref: 'v0.15.0'
 
       - uses: dtolnay/rust-toolchain@1.65.0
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+            SQLd ${{ github.ref }}" 
+          body_path: RELEASE_NOTES.md
+          prerelease: false
+          tag_name: ${{ github.ref_name }}
+          generate_release_notes: true
+          fail_on_unmatched_files: false
+          files: |
+             sqld-${{ github.ref }}-windows-latest/*
+             sqld-${{ github.ref }}-macos-latest/*
+             sqld-${{ github.ref }}-ubuntu-latest/*
```
MarinPostma, line 126: @dyasny the release notes are already taken care of by release-drafter: https://github.com/release-drafter/release-drafter, please do no override the body 🙏

dyasny: Does the drafter place a text file somewhere? I can simply point at it

dyasny: sure, I can do that. Maybe, since there's a process in place already, it's better to merge the two pipelines instead of having them? I can add the steps wherever they are most relevant and add conditions for the release to only get generated on a 'v*.*.*' tag creation for example. I assumed there wasn't anything in place because Glauber asked me to create this pipeline

MarinPostma: there isn't a workflow, the releases are drafted by the bot that is configured there: https://github.com/libsql/sqld/blob/main/.github/release-drafter.yml. When we are ready to cut a release, we bump the crate version and create the release from the github interface. This will generate a tag, and create the release.

PR #460, opened 2023-06-06
Testing setup for bottomless

Horusiath:
This PR is a PoC for possible environment that enables testing of various sold primitives, with bottomless backup/recovery process as an example.

While it requires a MinIO (S3 mockup service) running in a background, it's fairly lightweight - no containers or expensive process spawning. It can be run via `cargo test` and you can easily attach a debugger into it.

### DIFF
--- a/bottomless/test/minio-start.sh
+++ b/bottomless/test/minio-start.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+export AWS_ACCESS_KEY_ID=miniotest
+export AWS_DEFAULT_REGION=eu-west-1
+export AWS_SECRET_ACCESS_KEY=miniotest
+export LIBSQL_BOTTOMLESS_ENDPOINT=http://127.0.0.1:9000/
+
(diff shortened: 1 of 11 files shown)


bottomless/test/minio-start.sh
```diff
@@ -0,0 +1,13 @@
...
+export AWS_DEFAULT_REGION=eu-west-1
+export AWS_SECRET_ACCESS_KEY=miniotest
+export LIBSQL_BOTTOMLESS_ENDPOINT=http://127.0.0.1:9000/
+
+DIRECTORY=$(cd `dirname $0` && pwd)
+docker run -dt                                  \
+  -p 9000:9000 -p 9090:9090                     \
+  -v $DIRECTORY/minio-mnt/      \
+  --name "minio_local"                          \
+  quay.io/minio/minio server $DIRECTORY/minio-mnt/ --console-address ":9090"
```
Horusiath, line 13: This is an example starter script. We probably should configure GitHub Actions yaml file to start MinIO as well.

PR #468, opened 2023-06-08
disable nagle algorithm for http

MarinPostma:


penberg (approved):
Looks good! @honzasp, worth testing how much this reduces latency.

PR #469, opened 2023-06-08
Publish release binaries and docker images

dyasny:
Github action that publishes the sqld release binaries for linux and macosx x86

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,16 +1,17 @@
-name: Create and publish a Docker image
+name: Create and publish qld binaried and Docker image
 
 on:
   push:
     branches: ['main']
     tags:
-      - v*
+      - v*.*.*
 
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
+  # docker image build and upload to ghcr
   build-and-push-image:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,16 +1,17 @@
-name: Create and publish a Docker image
+name: Create and publish qld binaried and Docker image
```
MarinPostma, line 1:
```suggestion
name: Create and publish sqld binaries and Docker image
```

.github/workflows/publish.yml
```diff
@@ -43,3 +44,174 @@ jobs:
...
+      - uses: actions/checkout@v3
+
+      # Check Cargo.toml in sqld sbdir has the correct tag
+      - name: Verify sqld/Cargo.toml version is same as git tag
+        if: ${{ startsWith(github.ref, 'refs/tags/') }} && ${{ matrix.os }} == ubuntu-latest
+        id: check-tag
+        run: |
+          VER=$(echo ${{ github.ref_name }}|tr -d "v")
+          pushd sqld
+          CARGOVER=$(cargo metadata --no-deps | jq '.packages | .[] | select(.name == "sqld") | .version'|tr -d '"')
```
MarinPostma, line 72: will this work before you install a toolchain?

MarinPostma: also, is jq already installed?

dyasny: jq is there (see https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) but the order needs to be changed, I'll fix it

MarinPostma (approved):

PR #471, opened 2023-06-12
Improve sqld version reporting

penberg:
Include git hash and build date like `rustc`, for example, does, to make it easier to determine version of a `sqld` executable:

$ sqld --version
sqld 0.15.0 (acbce3dd 2023-06-12)

Useful when you have access to an executable, but not to the logs.

### DIFF
--- a/sqld/src/main.rs
+++ b/sqld/src/main.rs
@@ -9,7 +9,7 @@ use anyhow::{bail, Context as _, Result};
 use bytesize::ByteSize;
 use clap::Parser;
 use mimalloc::MiMalloc;
-use sqld::{database::dump::exporter::export_dump, Config};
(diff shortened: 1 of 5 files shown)


sqld/src/main.rs
```diff
@@ -21,7 +21,7 @@ static GLOBAL: MiMalloc = MiMalloc;
 /// SQL daemon
 #[derive(Debug, Parser)]
 #[command(name = "sqld")]
-#[command(about = "SQL daemon", version, long_about = None)]
+#[command(about = "SQL daemon", long_about = None)]
```
MarinPostma, line 24: https://docs.rs/clap/latest/clap/_derive/index.html#command-attributes `version` can take an expression parameter, no need to add a new argument

MarinPostma (approved):

PR #473, opened 2023-06-14
bottomless async replication

Horusiath:
This is proposal for async + batching backup in bottomless. Key points:

1. We batch sending multiple frames together. Every batch is a single S3 object, that has configurable number of frames inside. It supports compression.
2. Sending frames to S3 happens asynchronously in the background. Request to send frames is fully non-blocking, however atm. for consistency reasons we wait for confirmation until frames up to a particular number have been send.
3. There's no ongoing buffer copying. Replication progress is monitored using atomic counters. When batch send request is made, we copy entire WAL fragment into S3 PutObject request - this step is buffered but doesn't require full in-memory fragment copy.

TODOs:

- [x] make sync fully non-blocking - partially done, as sync itself is non-blocking, but on checkpoints we wait for frame backup confirmation.
- [x] remove existing ".consistent" object and rely only of frame batches. This will require detecting restored frames that are parts of non-finished transaction and postponing their merge into main DB file until transaction commit frame was received.
- [x] fix frames checksum verfication.
- [x] bind test suite to GitHub Actions.


bottomless/src/_replicator.rs
Horusiath, line 1: This is an old replicator implementation. I use it for reference. I'll remove it once we're done.

bottomless/src/wal.rs
```diff
@@ -0,0 +1,272 @@
+use crate::replicator::crc;
+use anyhow::{anyhow, Result};
+use std::io::{ErrorKind, SeekFrom};
+use std::path::Path;
+use tokio::io::{AsyncReadExt, AsyncSeekExt};
+
+const CRC_64: crc::Crc<u64> = crc::Crc::<u64>::new(&crc::CRC_64_ECMA_182);
+
+pub(crate) fn crc64(prev_crc: u64, data: &[u8]) -> u64 {
```
Horusiath, line 9: To be fixed with SQLite-compatible implementation: https://github.com/libsql/libsql/blob/f83d455ae8b57426b501770d392062d4f51165af/src/wal.c#L70-L88

bottomless/src/lib.rs
```diff
@@ -254,23 +256,25 @@ pub extern "C" fn xFrames(
             tracing::error!("{}", e);
             return ffi::SQLITE_IOERR_WRITE;
         }
-        for (pgno, data) in ffi::PageHdrIter::new(page_headers, page_size as usize) {
-            ctx.replicator.write(pgno, data);
-        }
+        let frame_count = ffi::PageHdrIter::new(page_headers, page_size as usize).count();
```
Horusiath, line 259: Any better way to get a number of frames to commit?

PR #474, opened 2023-06-14
Publish release

dyasny:
Now that we have an arm64 mac runner, adding the code to utilize it in release builds.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,16 +1,17 @@
-name: Create and publish a Docker image
+name: Create and publish qld binaried and Docker image
 
 on:
   push:
     branches: ['main']
     tags:
-      - v*
+      - v*.*.*
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -159,10 +158,10 @@ jobs:
             protobuf-compiler
 
       - name: install dependencies (macos)
-        if: matrix.os == 'macos-latest'
+        if: startsWith(matrix.os, 'macos')
         run: |
           brew install llvm
-          brew install pkg-config
+          brew install pkg-config<<<<<<< publish-release
```
MarinPostma, line 164: merge conflict?
```suggestion
          brew install pkg-config
```

MarinPostma (changes requested):

dyasny:
I also removed the on.push.["main"] bit, now it will only publish when a new v.*.*.* tag is created

MarinPostma (approved):

PR #497, opened 2023-07-03
Add --snapshot-exec to execute script for every generated snapshot file

honzasp:
Every time that sqld produces a snapshot file, the `--snapshot-exec` program will be executed with a path to this file. This provides a hook for external replication.

### DIFF
--- a/sqld/src/replication/primary/logger.rs
+++ b/sqld/src/replication/primary/logger.rs
@@ -12,6 +12,7 @@ use bytes::{Bytes, BytesMut};
 use parking_lot::RwLock;
 use sqld_libsql_bindings::init_static_wal_method;
 use tokio::sync::watch;
+use tokio::time::{Duration, Instant};
 use uuid::Uuid;
 
 #[cfg(feature = "bottomless")]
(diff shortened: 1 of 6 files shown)


MarinPostma:
Looking good, it's quite fragile though, since the script is not allowed to fail or the database shuts down.

Also, should we send merged snapshots as well?

honzasp:
> Looking good, it's quite fragile though, since the script is not allowed to fail or the database shuts down.

That's intentional: if the external script fails, it means that the external system may have lost the replicated data, which is a serious problem.

> Also, should we send merged snapshots as well?

I don't know :) It depends on how replication to the external system is designed.

honzasp:
We also need to be able to create snapshots periodically; I don't know how to do it, that's why this PR is still a draft.

MarinPostma:
> It's intentional..

Yeah, i know, maybe we should retry though, and require the script execution to be idempotent

sqld/src/replication/primary/logger.rs
```diff
@@ -863,6 +894,32 @@ impl ReplicationLogger {
...
+                return Ok(false)
+            };
+            last_frame_res?
+        };
+
+        let size_after = last_frame.header().size_after;
+        if size_after == 0 {
+            // last frame does not seem to be a commit boundary, so we can't do a compaction
+            return Ok(false);
+        }
```
MarinPostma, line 918: We can `assert` that: rev_iter's last frame must be a commit boundary since the iterator does not return uncommitted frames.

honzasp: Ok, changed

MarinPostma (approved):

PR #517, opened 2023-07-10
Make living replica keep primary alive

### LINKED ISSUE #506: Make replication requests count as idle kick
We currently only consider client requests as "idle kick" that defers scale to zero. However, a primary should also consider replication frame requests as "idle kick"; otherwise you might have a very active primary (that only serves replication requests) that scales to zero.

haaawk:
Each request for new frames from a replica to a primary resets the idle timer but such request is a long poll. It won't return until it gets some new frame.
This means that a replica won't send another request for new frames in the mean-time and won't restart the idle timer.

This commit makes primary keep track of replicas waiting for new frames and prevents IdleShutdownLayer from stopping the primary if there is at least one replica still waiting for the new frames.

When a replica is put into sleep, it will drop the connection and the connected_replicas counter will be decreased so only live replicas can prevent primary from going to sleep.

Fixes #506

### DIFF
--- a/sqld/src/rpc/mod.rs
+++ b/sqld/src/rpc/mod.rs
@@ -28,7 +28,7 @@ pub async fn run_rpc_server<D: Database>(
     idle_shutdown_layer: Option<IdleShutdownLayer>,
 ) -> anyhow::Result<()> {
     let proxy_service = ProxyService::new(factory, logger.new_frame_notifier.subscribe());
-    let logger_service = ReplicationLogService::new(logger);
+    let logger_service = ReplicationLogService::new(logger, idle_shutdown_layer.clone());
 
     tracing::info!("serving write proxy server at {addr}");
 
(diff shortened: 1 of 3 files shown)


haaawk:
FYI @athoscouto

penberg (commented):
Looks good to me, but @MarinPostma or @honzasp should review before merging.

haaawk:
> `FrameStream` should not know about the IdleShutdown logic. Here is what I propose:
> 
> ```rust
> struct StreamGuard<S> {
> s: S,
> guard: Arc<Guard>,
> }
> 
> impl Stream for StreamGuard {}
> struct Guard;
> 
> struct IdleShutdown {
> // ...
> guard: Arc<guard>,
> }
> 
> // to check if there still is a guard:
> isl.guard.strong_count() > 1
> ```

Ok. I will decouple the two. Thanks for the feedback @MarinPostma

MarinPostma (approved):

PR #528, opened 2023-07-13
Introduce initial-idle-shutdown-timeout-s

haaawk:
This new parameter makes it possible to define different
idle shutdown timeout for instance that has been started
but never received any requests.

It is useful for maintainance operations like periodic syncing
replica with the primary even when the replica does not serve
any requests at the moment. The assumption is that such maintainance
operation could wake up the replica for much shorter time than
idle-shutdown-timeout used after receiving actual user request.

### DIFF
--- a/sqld/src/main.rs
+++ b/sqld/src/main.rs
@@ -122,6 +122,12 @@ struct Cli {
     #[clap(long, env = "SQLD_IDLE_SHUTDOWN_TIMEOUT_S")]
     idle_shutdown_timeout_s: Option<u64>,
 
+    /// Like idle_shutdown_timeout_s but used only once after the server is started.
+    /// After that server either is shut down because it does not receive any requests
+    /// or idle_shutdown_timeout_s is used moving forward.
+    #[clap(long, env = "SQLD_INITIAL_IDLE_SHUTDOWN_TIMEOUT_S")]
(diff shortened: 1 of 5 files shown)


sqld/src/main.rs
```diff
@@ -122,6 +122,12 @@ struct Cli {
     #[clap(long, env = "SQLD_IDLE_SHUTDOWN_TIMEOUT_S")]
     idle_shutdown_timeout_s: Option<u64>,
 
+    /// Like idle_shutdown_timeout_s but used only once after the server is started.
+    /// After that server either is shut down because it does not receive any requests
+    /// or idle_shutdown_timeout_s is used moving forward.
+    #[clap(long, env = "SQLD_INITIAL_IDLE_SHUinitialTDOWN_TIMEOUT_S")]
```
MarinPostma, line 128:
```suggestion
    #[clap(long, env = "SQLD_INITIAL_IDLE_SHUTDOWN_TIMEOUT_S")]
```

MarinPostma (changes requested):

MarinPostma:
otherwise LGTM

sqld/src/main.rs
haaawk: Thanks for catching this.

MarinPostma (approved):

PR #534, opened 2023-07-19
sqld: Remove external `protoc` dependency

penberg:
Let's use the `protobuf-src` crate to remove the dependency to external `protoc` to simplify the build process.


haaawk (approved):
LGTM
I think you could also clean up https://github.com/libsql/sqld/blob/main/Dockerfile#L5

psarna (approved):

PR #549, opened 2023-07-25
Use cargo-chef

haaawk:
cargo-chef allows to split downloading and building dependencies from building the actual project. This allows docker to cache a layer that downloads and builds dependencies and rebuild only the project instead.

This change reduced the incremental build for me by over 60% from 327s to 130s.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,27 @@
 # build sqld
-FROM rust:slim-bullseye AS builder
+FROM rust:slim-bullseye AS chef
 RUN apt update \
     && apt install -y libclang-dev clang \
         build-essential tcl protobuf-compiler file \
         libssl-dev pkg-config git\
-    && apt clean
+    && apt clean \
+    && cargo install cargo-chef
+# We need to install and set as default the toolchain specified in rust-toolchain.toml
+# Otherwise cargo-chef will build dependencies using wrong toolchain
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,27 @@
...
+    && apt clean \
+    && cargo install cargo-chef
+# We need to install and set as default the toolchain specified in rust-toolchain.toml
+# Otherwise cargo-chef will build dependencies using wrong toolchain
+# This also prevents planner and builder steps from installing the toolchain over and over again
+COPY rust-toolchain.toml rust-toolchain.toml
+RUN cat rust-toolchain.toml | grep "channel" | awk '{print $3}' | sed 's/\"//g' > toolchain.txt \
+    && rustup update $(cat toolchain.txt) \
+    && rustup default $(cat toolchain.txt) \
+    && rm toolchain.txt rust-toolchain.toml
```
LucioFranco, line 16: Is this still needed? If the toolchain file is in the root of the cargo workspace it should pick it up?

haaawk: It is entirely possible. Let me check and get back to you :)

haaawk: Actually even if rust-toolchain.toml is picked up. Not running rustup update here will cause the toolchain to be installed on each step instead of installing it once so this way it's faster and cached better.

PR #565, opened 2023-07-30
Cleanup S3 bucket after bottomless test

neubaner:
Since we cannot rely on `S3BucketCleaner`'s drop guard to cleanup the S3 bucket after the test execution, I moved the cleanup into a function that takes a `Future` instead.

Also changed the bottomless options to use S3_URL instead of relying on configured environment variables.

### DIFF
--- a/sqld/src/test/bottomless.rs
+++ b/sqld/src/test/bottomless.rs
@@ -45,6 +45,7 @@ async fn backup_restore() {
                 bucket_name: BUCKET.to_string(),
                 max_batch_interval: Duration::from_millis(250),
                 restore_transaction_page_swap_after: 1, // in this test swap should happen at least once
+                aws_endpoint: Some(S3_URL.to_string()),
                 ..bottomless::replicator::Options::from_env().unwrap()
             }),
             db_path: PATH.into(),
@@ -233,6 +234,7 @@ async fn rollback_restore() {
                 bucket_name: BUCKET.to_string(),
                 max_batch_interval: Duration::from_millis(250),
                 restore_transaction_page_swap_after: 1, // in this test swap should happen at least once
+                aws_endpoint: Some(S3_URL.to_string()),
                 ..bottomless::replicator::Options::from_env().unwrap()
             }),
             db_path: PATH.into(),
--- a/sqld/src/test/bottomless.rs
+++ b/sqld/src/test/bottomless.rs
@@ -1,6 +1,7 @@
 use crate::{run_server, Config};
 use anyhow::Result;
 use libsql_client::{Connection, QueryResult, Statement, Value};
+use std::future::Future;
 use std::net::ToSocketAddrs;
 use std::path::PathBuf;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


sqld/src/test/bottomless.rs
```diff
@@ -374,49 +381,39 @@ impl Drop for DbFileCleaner {
...
-/// S3 bucket dir at the beginning and end of a test.
-struct S3BucketCleaner(&'static str);
-
-impl S3BucketCleaner {
-    async fn new(bucket: &'static str) -> Self {
-        let _ = Self::cleanup(bucket).await; // cleanup the bucket before test
-        S3BucketCleaner(bucket)
-    }
+/// Clean up the test data from S3 bucket dir at the beginning and end of a test.
+async fn with_s3_bucket_cleanup<T>(bucket: &'static str, fut: impl Future<Output = T>) -> T {
```
Horusiath, line 385: My main but hurt here is that it won't work when panics happen (which is frequent in tests due to assertions).

neubaner: You're right, I'll restore the guard and try something else :)

neubaner: @Horusiath I found two solutions using the guard:

The first one is to create a runtime on another thread. The reason `futures::executor::block_on` hanged was because it blocks the thread that tokio uses to drive their tasks.
```rust
impl Drop for S3BucketCleaner {
     fn drop(&mut self) {
         std::thread::spawn(|| {
             let _ = tokio::runtime::Builder::new_current_thread()
                 .enable_all()
                 .build()
                 .unwrap()
                 .block_on(Self::cleanup(self.0));
         })
         .join()
         .unwrap();
     }
}
```

The other solution is to use [`tokio::task::block_in_place`](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html) with a multi-thread scheduler:
```rust
impl Drop for S3BucketCleaner {
    fn drop(&mut self) {
        tokio::task::block_in_place(|| {
            let _  = tokio::runtime::Handle::current().block_on(Self::cleanup(self.0));
        });
    }
}
```
But it requires the tests to be run with:
```rust
#[tokio::test(flavor = "multi_thread")]
```

If any of those solutions satisfies you, I can update my PR 😄.

Thanks!

neubaner: @Horusiath I found two solutions using the guard:

The first one is to create a runtime on another thread. The reason futures::executor::block_on hanged was because it blocks the thread that tokio uses to drive their tasks.

```rust
impl Drop for S3BucketCleaner {
     fn drop(&mut self) {
         std::thread::spawn(|| {
             let _ = tokio::runtime::Builder::new_current_thread()
                 .enable_all()
                 .build()
                 .unwrap()
                 .block_on(Self::cleanup(self.0));
         })
         .join()
         .unwrap();
     }
}
```

The other solution is to use tokio::task::block_in_place with a multi-thread scheduler:
```rust
impl Drop for S3BucketCleaner {
    fn drop(&mut self) {
        tokio::task::block_in_place(|| {
            let _  = tokio::runtime::Handle::current().block_on(Self::cleanup(self.0));
        });
    }
}
```

But it requires the tests to be run with:
```rust
#[tokio::test(flavor = "multi_thread")]
```
If any of those solutions satisfies you, I can update my PR 😄.

Thanks!

Horusiath: I'd say that multi_thread scheduler might perform better given the fair amount to extra fibres created by sqld server and running in these tests.

gneubaner: Done!

neubaner: Done!

Horusiath (approved):
IMO it looks good to go.

PR #575, opened 2023-08-02
namespaces

MarinPostma:
This PR introduces namespaces to sqld.

Namespaces enable multiple database to be managed by a single sqld instance. A primary sqld instance can manage primary databases, and a replica sqld instance can manage replica databases.

Namespaces are lazily created based on the value of the host header. The host header must be present for every request. It is split by `.` and the first split determines the namespace. For example, if the host is `somedatabase.turso.com`, then the namespace is `somedatabase`. For websocket the handshake host header must also be set.

A hrana connection is opened for a specific namespace. In other words, a hrana connection can only open a connection to the database specified during the handshake.

Namespaces are lazily created on either the primary or a replica. When either receives a request to some namespaces, they create it on the fly. In the replica case, a replication stream is created on namespace creation that will trigger namespace creation on the primary. When a namespace is created on the primary, the creation is not eagerly propagated to the replicas. Rather, the replica will start replication from the primary upon the first request it gets for that namespace.

Integration with bottomless works seamlessly. The database id of the bottomless replicator is set to the namespace name, and the primary can lazily recover a database from bottomless if it exists there before responding to a request.

The snapshot script callback is also passed to the namespace. as a second argument

The `--allow-default-namespace` flag can be used to fall back to the `default` namespace if the host header can't be parsed properly.

`--allow-replica-overwrite` has been removed since the handshake is now explicit about the database we are trying to replicate. In case of log incompatibility, the replica is always reset.

## Remaining work:
- [ ] Finer-grained authentication: currently, the same token is used to query any database
- [ ] Loading dumps: We need an API to load dumps into a namespace
- [x] ~Create dump: we need an API to perform a dump for a namespace~ dumps can be created for specific namespaces by passing the `--namespace` arguments to the dump subcommand.
- [ ] Eager namespace creation: namespaces are lazily created. Let's move to a more convenient model where namespaces are created on demand through the admin API
- [ ] Allow namespace deletion: deleting namespaces can be tricky since we need to ensure consistency between replicas and the primary


penberg:
> Finer-grained authentication: currently, the same token is used to query any database

This is fine for now. We can merge without this, let's just open a ticket to remind us that we need to do it.

sqld/src/hrana/ws/handshake.rs
```diff
@@ -14,14 +19,29 @@ pub enum WebSocket {
     Upgraded(tokio_tungstenite::WebSocketStream<hyper::upgrade::Upgraded>),
 }
 
-pub async fn handshake_tcp(socket: tokio::net::TcpStream) -> Result<(WebSocket, Version)> {
+fn extract_namespace<B>(req: &Request<B>) -> anyhow::Result<Bytes> {
+    let host = req.headers().get("host").context("missing host header")?;
```
LucioFranco, line 23: I think we should be avoiding to use context and convert this directly into its error type that is well defined and we can then look at how it implements IntoResponse to understand what status/message we will reply with.

sqld/src/http/mod.rs
```diff
@@ -243,7 +258,8 @@ pub async fn run_http<D: Database>(
         tracing::debug!("got request: {} {}", req.method(), req.uri());
     }
 
-    let app = Router::new()
+    let app = Router::new();
+    let app = app
```
LucioFranco, line 262: why this change?

sqld/src/rpc/replication_log.rs
```diff
@@ -125,15 +127,29 @@ impl ReplicationLog for ReplicationLogService {
...
         }
 
+        let logger = match self
+            .namespaces
+            .with(req.namespace, |ns| ns.meta.logger.clone())
+            .await
+        {
+            Ok(logger) => logger,
+            Err(e) => {
+                return Err(Status::internal(format!(
```
LucioFranco, line 145: I believe we could do the same for status and have conversion code in error.rs?

sqld/src/namespace.rs
```diff
@@ -0,0 +1,304 @@
...
+use crate::replication::replica::Replicator;
+use crate::replication::{NamespacedSnapshotCallback, ReplicationLogger};
+use crate::stats::Stats;
+use crate::{
+    check_fresh_db, init_bottomless_replicator, run_periodic_compactions, DB_CREATE_TIMEOUT,
+    MAX_CONCURRENT_DBS,
+};
+
+#[async_trait::async_trait]
+pub trait NamespaceFactory: Sync + Send + 'static {
```
LucioFranco, line 29: I think this trait name is super confusing. I would maybe call it `NamedspacedDb` and just drop the factory since I think we all would know this is a factory :D (or can call it Maker which feels more rusty).

sqld/src/namespace.rs
```diff
@@ -0,0 +1,304 @@
...
+    type Database = TrackedDb<WriteProxyDatabase>;
+    type Meta = ();
+
+    async fn create(&self, name: Bytes) -> anyhow::Result<Namespace<Self::Database, Self::Meta>> {
+        Namespace::new_replica(&self.config, name).await
+    }
+}
+
+pub struct Namespaces<F: NamespaceFactory> {
+    #[allow(clippy::type_complexity)]
```
LucioFranco, line 77: We should just globally allow this lint....its really dumb

sqld/src/namespace.rs
```diff
@@ -0,0 +1,304 @@
...
+            let ns = self.factory.create(namespace.clone()).await?;
+            let ret = f(&ns);
+            lock.insert(namespace, ns);
+            Ok(ret)
+        }
+    }
+}
+
+#[allow(dead_code)]
+pub struct Namespace<D, T> {
```
LucioFranco, line 126: This name as well doesn't make sense to me either would be good to have a doc comment or rename it to something more digestible.

sqld/src/namespace.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub extensions: Vec<PathBuf>,
+    pub stats: Stats,
+    pub config_store: Arc<DatabaseConfigStore>,
+    pub max_response_size: u64,
+    pub max_total_response_size: u64,
+    pub hard_reset: mpsc::Sender<Bytes>,
+}
+
+#[async_trait::async_trait]
+pub trait NamespaceCommon {
```
LucioFranco, line 146: Should this just be called `Namespace` and we rename the struct above?

sqld/src/http/mod.rs
```diff
@@ -193,48 +195,61 @@ async fn handle_fallback() -> impl IntoResponse {
...
+    namespaces: Arc<Namespaces<F>>,
     upgrade_tx: mpsc::Sender<hrana::ws::Upgrade>,
-    hrana_http_srv: Arc<hrana::http::Server<D>>,
+    hrana_http_srv: Arc<hrana::http::Server<F::Database>>,
     enable_console: bool,
     idle_shutdown_layer: Option<IdleShutdownLayer>,
     stats: Stats,
-    logger: Option<Arc<ReplicationLogger>>,
-) -> anyhow::Result<()> {
+    replication_service: Option<S>,
```
LucioFranco, line 233: I couldn't figure out why this was changed to be generic? Could you expand on that, I think adding generics if we don't need it can make this API much harder to read.

LucioFranco (commented):
Overall looks really good! I think we could use a better PR title and I left a few nits/comments.

sqld/src/http/mod.rs
MarinPostma: no reason, I was struggling with axum type error, and forgot to change it back :)

sqld/src/hrana/ws/handshake.rs
```diff
@@ -14,14 +19,29 @@ pub enum WebSocket {
...
+    let mut namespace = None;
     let callback = |req: &http::Request<()>, resp: http::Response<()>| {
         let (mut resp_parts, _) = resp.into_parts();
         resp_parts
             .headers
             .insert("server", http::HeaderValue::from_static("sqld-hrana-tcp"));
 
+        namespace = match extract_namespace(req) {
+            Ok(ns) => Some(ns),
+            Err(_) if allow_default_ns => Some(DEFAULT_NAMESPACE_NAME.into()),
```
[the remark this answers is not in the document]
MarinPostma: I don't think you can do that since `namespace` is set by the callback body, and Rust can't prove that the callback will be invoked at any point, leaving the namespace uninitialized. Or maybe I misunderstood your suggestion?

sqld/src/http/mod.rs
MarinPostma: 2 step plan:
- don't leak namespace-specific information to HTTP (`Namespace<PrimaryNamespaceFactory>`)
- rn, only the primary can serve replication requests, but, it's easy enough to make a proxy rpc service so that replication request can be served from any node, which is what I want to do in a followup PR.

MarinPostma:
@LucioFranco optionally [a026a89](https://github.com/libsql/sqld/pull/575/commits/a026a89d750ad64fe590031805bab937669f0e69) contains a refactor that should help with clarity. I can split it off in another PR if necessary, lmk.

I tried to make the concepts a bit clearer between database/connection/namespace

sqld/src/http/mod.rs
LucioFranco: Would it be better to just enum that service rather than make things more generic?

sqld/src/rpc/replication_log.rs
```diff
@@ -8,23 +8,25 @@ use std::net::SocketAddr;
...
 use crate::utils::services::idle_shutdown::IdleShutdownLayer;
 
 use self::rpc::replication_log_server::ReplicationLog;
 use self::rpc::{Frame, Frames, HelloRequest, HelloResponse, LogOffset};
 
 pub struct ReplicationLogService {
-    logger: Arc<ReplicationLogger>,
-    replicas_with_hello: RwLock<HashSet<SocketAddr>>,
+    namespaces: Arc<NamespaceStore<PrimaryNamespaceMaker>>,
+    replicas_with_hello: RwLock<HashSet<(SocketAddr, Bytes)>>,
```
LucioFranco, line 29: Is this bytes supposed to be a namespace? If so I would suggest we maybe newtype this to make sure its clear that this bytes was extracted as a namespace, this can be done in a follow up though.

LucioFranco (commented):
Everything looks really good, we chatted about some naming stuff over slack but no need to fix them now. Almost all my comments are stuff we can do in a follow up so if we don't do them now we should at least make an issue that captures all of the changes we might want to make in the future.

The only comment I want to bring to your attention is the one around the namespace extraction and the fly domain, I think that is a blocker for at least getting embedded replicas somewhat working due to the fly issue. Let me know if you have any comments.

sqld/src/http/db_factory.rs
```diff
@@ -0,0 +1,64 @@
...
+        Ok(Self(
+            state
+                .namespaces
+                .with(ns, |ns| ns.db.connection_maker())
+                .await?,
+        ))
+    }
+}
+
+pub fn namespace_from_headers(
```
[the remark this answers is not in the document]
LucioFranco: This is actually not an issue since we pass namespace via protobuf message and that is where this issue with domains is. So this is no longer a problem.

LucioFranco (approved):
:shipit: 🔥

PR #577, opened 2023-08-04
intit bottomless before replication log recovery

MarinPostma:
This pr moves the initialization of bottomless before the replication log is initialized. If a fresh database needs to recover from bottomless we want the replication log to be regenerated, so it must be initialized after the database has been restored.

Also, if the database has been recovered by bottomless, we must regenerate the replication log.


MarinPostma:
superseded by #575

PR #584, opened 2023-08-07
sqld: Add `BatchLogEntries` and replication rpc auth

LucioFranco:
This PR adds a new rpc endpoint that doesn't stream log entries fit for libsql and additionally adds auth to the replication service rpc code which is only enabled for the public facing http server and not the internal one.

### DIFF
--- a/sqld/proto/replication_log.proto
+++ b/sqld/proto/replication_log.proto
@@ -27,6 +27,6 @@ message Frames {
 service ReplicationLog {
     rpc Hello(HelloRequest) returns (HelloResponse) {}
     rpc LogEntries(LogOffset) returns (stream Frame) {}
-    rpc LogEntriesSnapshot(LogOffset) returns (Frames) {}
+    rpc BatchLogEntries(LogOffset) returns (Frames) {}
     rpc Snapshot(LogOffset) returns (stream Frame) {}
 }
--- a/sqld/proto/replication_log.proto
+++ b/sqld/proto/replication_log.proto
@@ -20,8 +20,13 @@ message Frame {
     bytes data = 1;
 }
 
+message Frames {
+    repeated Frame frames = 1;
+}
+
 service ReplicationLog {
     rpc Hello(HelloRequest) returns (HelloResponse) {}
     rpc LogEntries(LogOffset) returns (stream Frame) {}
+    rpc LogEntriesSnapshot(LogOffset) returns (Frames) {}
     rpc Snapshot(LogOffset) returns (stream Frame) {}
 }
--- a/sqld/src/replication/primary/frame_stream.rs
+++ b/sqld/src/replication/primary/frame_stream.rs
@@ -15,16 +15,22 @@ pub struct FrameStream {
     pub(crate) max_available_frame_no: FrameNo,
     logger: Arc<ReplicationLogger>,
     state: FrameStreamState,
+    wait_for_more: bool,
 }
 
 impl FrameStream {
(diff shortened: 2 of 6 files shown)


sqld/proto/replication_log.proto
```diff
@@ -20,8 +20,13 @@ message Frame {
...
 }
 
+message Frames {
+    repeated Frame frames = 1;
+}
+
 service ReplicationLog {
     rpc Hello(HelloRequest) returns (HelloResponse) {}
     rpc LogEntries(LogOffset) returns (stream Frame) {}
+    rpc LogEntriesSnapshot(LogOffset) returns (Frames) {}
```
haaawk, line 30: I'm not sure if Snapshot is a good name since we're using that name already for different concept.

haaawk: Maybe something like LogEntriesNonblocking would better describe the difference from LogEntries?

sqld/src/rpc/replication_log.rs
```diff
@@ -94,7 +95,7 @@ impl<S: futures::stream::Stream + Unpin> futures::stream::Stream for StreamGuard
         self: std::pin::Pin<&mut Self>,
         cx: &mut std::task::Context<'_>,
     ) -> std::task::Poll<Option<Self::Item>> {
-        self.get_mut().s.poll_next_unpin(cx)
+        Pin::new(&mut self.get_mut().s).poll_next(cx)
```
haaawk, line 111: Out of curiosity, what is this change doing?

sqld/src/rpc/replication_log.rs
```diff
@@ -34,11 +36,22 @@ impl ReplicationLogService {
...
+            auth,
+        }
+    }
+
+    fn authenticate<T>(&self, req: &tonic::Request<T>) -> Result<(), Status> {
+        if let Some(auth) = &self.auth {
+            let _ = auth.authenticate_grpc(req)?;
+            Ok(())
+        } else {
+            Ok(())
```
haaawk, line 54: Ultra nit: maybe:
```
if let Some(auth) = &self.auth {
  let _ = auth.authenticate_grpc(req)?;
}
Ok(())
```
?

haaawk (approved):

sqld/proto/replication_log.proto
haaawk: or maybe LogEntriesBatch?

sqld/src/rpc/replication_log.rs
LucioFranco: I removed the `futures` crate and moved to tokio_stream (which is better but happy to elaborate on slack). The only issue is that tokio_stream's ext trait doesn't have the unpin and boxed fn so I replaced those usages. This just does the same thing as the fn does but done manually.

sqld/src/replication/primary/frame_stream.rs
```diff
@@ -84,6 +90,12 @@ impl Stream for FrameStream {
                 }
 
                 Err(LogReadError::Ahead) => {
+                    // If we don't wait to wait for more then lets end this stream
+                    // without subscribing for more frames
+                    if !self.wait_for_more {
+                        return Poll::Ready(None);
```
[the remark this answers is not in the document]
LucioFranco: Yup, good catch, Ive added that as a new line.

psarna (approved):

sqld/src/rpc/replication_log.rs
haaawk: Cool. No objections. I was just curious. Thanks

PR #588, opened 2023-08-08
sqld: Add panic logging, api error and fix log env

LucioFranco:


sqld/src/main.rs
```diff
@@ -336,6 +336,10 @@ fn enable_libsql_logging() {
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    if std::env::var("RUST_LOG").is_err() {
```
LucioFranco, line 339: I couldn't get RUST_LOG to work locally...so I am doing it this way and it seems to work better. Idk why the previous setup was not working.

MarinPostma (approved):

psarna (approved):

PR #589, opened 2023-08-10
Create namespace from dump

MarinPostma:
This PR enables creating namespaces by loading from a sqlite dump.

To create a namespace from a dump, send the dump to `POST /v1/namespaces/:namespace_name/create-with-dump` on the admin API.

depends on #575


sqld/src/admin_api.rs
```diff
@@ -1,25 +1,37 @@
 use anyhow::Context as _;
-use axum::{extract::State, Json};
-use serde::Deserialize;
-use std::net::SocketAddr;
+use axum::extract::{BodyStream, Path, State};
+use axum::Json;
+use futures::TryStreamExt;
```
LucioFranco, line 4: I would avoid using futures crate and use tokio-stream instead

LucioFranco (approved):

sqld/src/admin_api.rs
MarinPostma: Why is that?

PR #594, opened 2023-08-10
fix(cargo-metadata): update bottomless-cli repo

ashleygwilliams:
👋 a very small change that i noticed while trying out cargo-dist on this repo


MarinPostma:
Thanks!

PR #599, opened 2023-08-11
proxy frame replication

MarinPostma:
This PR adds the ability to proxy replication request from replica to its primary. Essentially, now, replication requests can hit any instance in a sqld cluster.


penberg:
Very cool, @MarinPostma! I will let @LucioFranco review this.

honzasp (approved):

PR #613, opened 2023-08-21
return appropriate sqlite error when checkpoint it ignored

MarinPostma:
the `on_checkpoint` hook inhibits checkpointing if the checkpoint flag in anything less than `TRUNCATE`. The hook was returning `SQLITE_OK` in those situation. SQLite interprets `xCheckpoint` returning `SQLITE_OK` as the WAL being completely checkpointed and the database file being `fsync`ed. On close, it then assumes that the WAL is empty, and simply detroys it, resulting in data losses.

Returning `SQLITE_BUSY` instead prevents that.

sources:
- [code that invoked the wal method](https://github.com/libsql/libsql/blob/main/src/wal.c#L2258-L2273)
- [sqlite returning `SQLITE_BUSY` when failing to checkpoint](https://github.com/libsql/libsql/blob/main/src/wal.c#L3912)

Since we don't really check if the checkpoint was triggered by a connection close, a connection close never deletes the WAL file, because a connection close always uses the `PASSIVE` checkpoint flag. This should not be an issue, and leaves us full control over checkpointing.

### DIFF
--- a/sqld/src/replication/primary/logger.rs
+++ b/sqld/src/replication/primary/logger.rs
@@ -10,6 +10,7 @@ use anyhow::{bail, ensure};
 use bytemuck::{bytes_of, pod_read_unaligned, Pod, Zeroable};
 use bytes::{Bytes, BytesMut};
 use parking_lot::RwLock;
+use rusqlite::ffi::SQLITE_BUSY;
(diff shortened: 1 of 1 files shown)


Horusiath (approved):
LGTM

PR #614, opened 2023-08-21
bottomless-cli: added support for namespaces

Horusiath:
This PR fixes `bottomless-cli` tool to support namespaces added some time ago (see #575). Changes include:
1. Changed bottomless S3 keyspace prefix: atm its `ns-{namespace}data-{uuid}/`. With this PR it will be: `ns-{namespace}:{db_path}-{uuid}/`. This way we can perform DB recovery directly to its original path and setup required options.
2. `bottomless-cli ls` now correctly infers database path from S3 bucket - previously it was option, but never actually worked.
3. `bottomless-cli ls` now displays not only generations but also their creation times (useful since bottomless supports point-in-time recovery for some time).
4. Point-in-time recovery was broken by #566 (basically library we use was no longer able to decode timestamp from v7 UUIDs we created) and was silently failing. This PR also fixes that.

### DIFF
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -1,6 +1,7 @@
 use crate::backup::WalCopier;
 use crate::read::BatchReader;
 use crate::transaction_cache::TransactionPageCache;
+use crate::uuid_utils::decode_unix_timestamp;
 use crate::wal::WalFileReader;
 use anyhow::anyhow;
 use arc_swap::ArcSwap;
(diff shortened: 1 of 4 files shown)


haaawk:
@MarinPostma could you please confirm that the namespace handling is correct here?

MarinPostma:
I'm wondering about making bottomless namespace-aware. Do we really want that? The only thing that sqld does is to set the db id, maybe it would be sufficient to pass that as a cli arg, and let bottomless-cli be oblivious of what namespaces are?

bottomless/src/replicator.rs
```diff
@@ -239,12 +240,13 @@ impl Replicator {
         let db_path = db_path.into();
         let db_name = {
             let db_id = options.db_id.unwrap_or_default();
-            let name = match db_path.rfind('/') {
-                Some(index) => &db_path[index + 1..],
+            let name = match db_path.find('/') {
+                Some(index) => &db_path[..index],
```
Horusiath, line 244: This change is because previously we used the last database path segment, which was always `/data`. I've changed that to use first segment, but it may have it's own issues depending on what will we provide as a path.

Any thoughts @psarna ?

MarinPostma (approved):

PR #618, opened 2023-08-24
Migrate database directory structure from pre 0.18

MarinPostma:
This PR introduce automatic detection and migration of of imcompatible version of sqld.

For now, only migration from pre 0.18 directory structure to the new directory structure is supported.

### DIFF
--- a/sqld/src/migration.rs
+++ b/sqld/src/migration.rs
@@ -1,5 +1,5 @@
-use std::path::Path;
 use std::fs::read_to_string;
+use std::path::Path;
 
 use anyhow::Context;
 use semver::Version as SemVer;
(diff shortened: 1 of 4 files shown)


athoscouto (approved):

sqld/src/migration.rs
```diff
@@ -0,0 +1,87 @@
...
+        let ns_dir = db_path.join("dbs").join("default");
+        std::fs::create_dir_all(&ns_dir)?;
+
+        macro_rules! maybe_link {
+            ($name:expr) => {
+                if db_path.join($name).exists() {
+                    std::fs::hard_link(db_path.join($name), ns_dir.join($name))?;
+                }
+            };
+        }
```
LucioFranco, line 49: This could just be a fn and not a macro, it would make this easier to read.

sqld/src/migration.rs
```diff
@@ -0,0 +1,87 @@
...
+        let _ = std::fs::remove_dir_all(db_path.join("dbs"));
+        return Err(e);
+    }
+
+    // best effort cleanup 
+    let _ = std::fs::remove_file(db_path.join("data"));
+    let _ = std::fs::remove_file(db_path.join("data-shm"));
+    let _ = std::fs::remove_file(db_path.join("data-wal"));
+    let _ = std::fs::remove_file(db_path.join("wallog"));
+    let _ = std::fs::remove_file(db_path.join("client_wal_index"));
```
LucioFranco, line 84: I think we should log these at least just incase something weird happens.

LucioFranco (approved):

PR #623, opened 2023-08-25
eager namespace creation

MarinPostma:
This PR changes the way namespaces are created from lazy to eager. Namespaces need to be explicitely created on the primary using the `POST /v1/namespaces/:namespace/create` or `POST /v1/namespaces/:namespace/create-with-dump` routes on the admin API.

It works by disabling lazy creation on the primary, so that namespaces can only be created explicitely there, but still allowing the lazy creation on replicas. When a namespaces exists on the primary but not on a replica, and a request hits the replica, the replica will create a namespace and try to replicate from it. On the other hand if the replica doesn't exist, the primary will return an error stating that the namespaces does not exist, and the replica will abort it's namespace creation.

### DIFF
--- a/sqld/src/error.rs
+++ b/sqld/src/error.rs
@@ -55,6 +55,8 @@ pub enum Error {
     InvalidHost(String),
     #[error("namespace `{0}` doesn't exist")]
     UnexistingNamespace(String),
+    #[error("namespace `{0}` already exists")]
+    NamespaceAlreadyExist(String),
     #[error("replication error: {0}")]
     ReplicationError(#[from] ReplicationError),
     #[error("Failed to connect to primary")]
@@ -99,6 +101,7 @@ impl IntoResponse for Error {
             UnexistingNamespace(_) => self.format_err(StatusCode::BAD_REQUEST),
             ReplicationError(_) => self.format_err(StatusCode::INTERNAL_SERVER_ERROR),
             PrimaryConnectionTimeout => self.format_err(StatusCode::INTERNAL_SERVER_ERROR),
+            NamespaceAlreadyExist(_) => self.format_err(StatusCode::BAD_REQUEST),
         }
     }
 }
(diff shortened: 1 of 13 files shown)


sqld/src/error.rs
```diff
@@ -53,6 +56,16 @@ pub enum Error {
     Anyhow(#[from] anyhow::Error),
     #[error("Invalid host header: `{0}`")]
     InvalidHost(String),
+    #[error("Namespace `{0}` doesn't exist")]
+    UnexistingNamespace(String),
```
LucioFranco, line 60:
```suggestion
    NamespaceDoesntExist(String),
```

This reads a bit better

sqld/src/namespace.rs
```diff
@@ -85,12 +100,24 @@ impl MakeNamespace for ReplicaNamespaceMaker {
         &self,
         name: Bytes,
         dump: Option<DumpStream>,
-    ) -> anyhow::Result<Namespace<Self::Database>> {
+        allow_creation: bool,
+    ) -> crate::Result<Namespace<Self::Database>> {
         if dump.is_some() {
-            bail!("cannot load dump on replica");
+            // bail!("cannot load dump on replica");
+            todo!()
```
LucioFranco, line 107: Whats the TODO here about?

sqld/src/rpc/mod.rs
```diff
@@ -15,6 +15,9 @@ pub mod proxy;
 pub mod replication_log;
 pub mod replication_log_proxy;
 
+/// A tonic error code to signify that a namespace doesn't exist.
+pub const UNEXISTING_NAMESPACE: &str = "NEED_SNAPSHOT";
```
LucioFranco, line 19: Should change this again with the wording we use for the error variant.

LucioFranco (changes requested):
Overall LGTM just a few adjustments then we can merge.

LucioFranco (approved):

PR #625, opened 2023-08-25
feat(dist): add cargo-dist

ashleygwilliams:


glommer:
Thanks 👑 !! @MarinPostma , FYI

MarinPostma:
Looks great! 
We should drop the part that creates the release. The releases are automatically drafted as we merge PRs; the tag is created when we publish the release. I think this recipe should be able to pick this up and add the artifacts to the newly created release.

ashleygwilliams:
makes sense @MarinPostma - will update shortly

MarinPostma (approved):
amazing, let's try it :) Thanks @ashleygwilliams

ashleygwilliams:
@MarinPostma our biggest question is how exactly the release-drafter names its draft releases and if our script will properly "find the draft". based on reading your [current publish workflow](https://github.com/libsql/sqld/blob/main/.github/workflows/publish.yml#L221) my suspicion is that it should (as both use the triggering tag), but i'll admit that how release-drafter does this is a bit unclear to me!

MarinPostma:
@ashleygwilliams just tried to release, but it seems to be broken: https://github.com/libsql/sqld/actions/runs/5979890643

ashleygwilliams:
ah yeah, shoot. i updated the job name but forgot to in other places. i'll fix.

PR #626, opened 2023-08-25
fix(dist): revert partial workflow job name change

ashleygwilliams:
in a rush to manually adapt the generated ci file, i only partially renamed the job i edited. this PR reverts it to it original name, which is a bit unfortunate as it refers primarily to thing i edited it to not do (create the release). regardless, i am also testing this on a fork to ensure it is valid and will report back.

we have a fix to avoid the manual edit that will be available in a pre-release monday.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,7 +46,7 @@ on:
 
 jobs:
   # Create the Github Release™ so the packages have something to be uploaded to
-  cargo-dist-plan:
+  create-release:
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
test run is here: https://github.com/ashleygwilliams/sqld-cargodist-demo/actions/runs/5980087663

MarinPostma (approved):

MarinPostma:
Thanks @ashleygwilliams 👍

PR #629, opened 2023-08-28
Namespace deletion

MarinPostma:
enable namespace deletion with the `DELETE /v1/namespaces/:namespaces` on the primary admin API.

Deleting a namespace on the primary releases all resources associated with that namespace, and removes the namespace directory and the backup from bottomless.

Once the namespace is removed from the primary, any replica attempting to replicate will get a `namespace_doesnt_exist` error, and will in turn remove all resources associated with that namespace.

### DIFF
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -14,7 +14,6 @@ use aws_sdk_s3::primitives::ByteStream;
 use aws_sdk_s3::{Client, Config};
 use bytes::{Buf, Bytes, BytesMut};
 use chrono::{DateTime, LocalResult, NaiveDateTime, TimeZone, Utc};
-use tokio::task::JoinSet;
 use std::io::SeekFrom;
 use std::ops::Deref;
 use std::path::Path;
@@ -23,6 +22,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt};
 use tokio::sync::watch::{channel, Receiver, Sender};
+use tokio::task::JoinSet;
 use tokio::time::{timeout_at, Instant};
 use uuid::{NoContext, Uuid};
 
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -14,6 +14,7 @@ use aws_sdk_s3::primitives::ByteStream;
 use aws_sdk_s3::{Client, Config};
 use bytes::{Buf, Bytes, BytesMut};
 use chrono::{DateTime, LocalResult, NaiveDateTime, TimeZone, Utc};
+use tokio::task::JoinSet;
 use std::io::SeekFrom;
 use std::ops::Deref;
 use std::path::Path;
@@ -54,6 +55,7 @@ pub struct Replicator {
     use_compression: CompressionKind,
     max_frames_per_batch: usize,
     s3_upload_max_parallelism: usize,
+    _join_set: JoinSet<()>,
 }
 
 #[derive(Debug)]
@@ -260,6 +262,8 @@ impl Replicator {
         let last_sent_frame_no = Arc::new(AtomicU32::new(0));
         let commits_in_current_generation = Arc::new(AtomicU32::new(0));
 
+        let mut _join_set = JoinSet::new();
+
         let (frames_outbox, mut frames_inbox) = tokio::sync::mpsc::channel(64);
         let _local_backup = {
             let mut copier = WalCopier::new(
@@ -274,7 +278,7 @@ impl Replicator {
             let next_frame_no = next_frame_no.clone();
             let last_sent_frame_no = last_sent_frame_no.clone();
             let batch_interval = options.max_batch_interval;
-            tokio::spawn(async move {
+            _join_set.spawn(async move {
                 loop {
                     let timeout = Instant::now() + batch_interval;
                     let trigger = match timeout_at(timeout, flush_trigger_rx.changed()).await {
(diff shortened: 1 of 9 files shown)


sqld/src/replication/primary/frame_stream.rs
```diff
@@ -13,25 +13,27 @@ use crate::replication::{FrameNo, LogReadError, ReplicationLogger};
...
-    logger: Arc<ReplicationLogger>,
+    logger: Weak<ReplicationLogger>,
     state: FrameStreamState,
     wait_for_more: bool,
 }
 
 impl FrameStream {
     pub fn new(
-        logger: Arc<ReplicationLogger>,
+        logger: Weak<ReplicationLogger>,
```
LucioFranco, line 23: I don't really like using weak I think it opens up a lot of opportunity to run into issues (and Ive seen it happen plenty in the past). Can we write this in such a way without weak? Or atleast can we just do all of the arc stuff within the logger so we don't expose this everywhere?

LucioFranco (changes requested):

sqld/src/namespace.rs
```diff
@@ -121,13 +140,37 @@ impl MakeNamespace for ReplicaNamespaceMaker {
...
-    factory: F,
+    make_namespace: M,
     allow_lazy_creation: bool,
 }
 
+impl<M: MakeNamespace> NamespaceStore<M> {
+    pub async fn destroy(&self, namespace: Bytes) -> crate::Result<()> {
+        let mut lock = self.inner.write().await;
+        if let Some(ns) = lock.remove(&namespace) {
+            // FIXME: when destroying, we are waiting for all the tasks associated with the
```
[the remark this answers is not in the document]
MarinPostma: We could, but I think I'll just move to use https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html

bottomless/src/replicator.rs
```diff
@@ -54,6 +55,7 @@ pub struct Replicator {
     use_compression: CompressionKind,
     max_frames_per_batch: usize,
     s3_upload_max_parallelism: usize,
+    _join_set: JoinSet<()>,
```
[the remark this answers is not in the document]
MarinPostma: I don't think we do  🤔

sqld/src/replication/primary/frame_stream.rs
MarinPostma: yeah, maybe this is a bit too brittle. Let me come up with something better

bottomless/src/replicator.rs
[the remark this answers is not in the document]
LucioFranco: 🤔 diff hard to read I guess lol

sqld/src/namespace.rs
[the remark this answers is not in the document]
LucioFranco: Cool should open an issue for this to track 😄

bottomless/src/replicator.rs
```diff
@@ -260,6 +262,8 @@ impl Replicator {
         let last_sent_frame_no = Arc::new(AtomicU32::new(0));
         let commits_in_current_generation = Arc::new(AtomicU32::new(0));
 
+        let mut _join_set = JoinSet::new();
```
LucioFranco, line 265: Ah this one we use?? I see that we use `_join_set` below??

sqld/src/lib.rs
```diff
@@ -333,12 +338,19 @@ async fn start_replica(
...
-                    "received reset signal for: {:?}",
-                    std::str::from_utf8(&ns).ok()
-                );
-                namespaces.reset(ns).await?;
+            while let Some(op) = hard_reset_rcv.recv().await {
+                match op {
+                    ResetOp::Reset(ns) => {
+                        tracing::warn!(
+                            "received reset signal for: {:?}",
+                            std::str::from_utf8(&ns).ok()
```
LucioFranco, line 346: can always do `String::from_utf8_lossy` that doesn't error.

LucioFranco (approved):

sqld/src/admin_api.rs
```diff
@@ -101,7 +103,15 @@ async fn handle_create_namespace_with_dump<F: MakeNamespace>(
...
+) -> crate::Result<()> {
     app_state.namespaces.create(namespace.into(), None).await?;
     Ok(())
 }
+
+async fn handle_delete_namespace<F: MakeNamespace>(
+    State(app_state): State<Arc<AppState<F>>>,
+    Path(namespace): Path<String>,
+) -> crate::Result<()> {
+    app_state.namespaces.destroy(namespace.into()).await?;
```
Horusiath, line 115: I've noticed that while we convert namespace to `Bytes` here, deeper in the call stack it's sometimes used as `String` again ie. when performing potentially non-safe UTF-8 conversion in order to include namespace in error message.

Maybe it would be better for these methods to take `String` as a parameter and only convert it to `Bytes` when necessary?

sqld/src/lib.rs
```diff
@@ -305,6 +305,11 @@ fn configure_rpc(config: &Config) -> anyhow::Result<(Channel, tonic::transport::
     Ok((channel, uri))
 }
 
+pub enum ResetOp {
+    Reset(Bytes),
+    Destroy(Bytes),
```
Horusiath, line 310: Why is it called `ResetOp`? Is destroying the database considered special case of reset? Are there any future operations that could eventually use the same mechanism this enum is used in?

Horusiath (approved):
LGTM, just left few questions.

sqld/src/admin_api.rs
MarinPostma: Initially I wanted it to be an Arc<str> because we keep cloning it everywhere, but this is not supported by tonic, so I had to resolve to use bytes instead

sqld/src/lib.rs
MarinPostma: There are two cases: when a replica is incompatible it is reset to replicate again, when a namespace doesn't exist, it is destroyed completely. This is the adaptation of old code, and it could probably use a refactor

PR #634, opened 2023-08-29
chore: update cargo-dist to 0.2.0-prerelease.6 and set create-release=false

Gankra:
This enables the builtin support for the functionality that was temporarily hand-edited into release.yml. prerelease.6 also incidentally got a bunch of other improvements, like support for more tag formats and cleaned up comments/terms in the release.yml.

(All I did was hand-edit in `create-release=false` and then ran `cargo dist init --yes` with `prerelease.6` installed on my computer).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,53 +4,50 @@
 # CI that:
 #
 # * checks for a Git Tag that looks like a release
-# * creates a Github Release™ and fills in its text
-# * builds artifacts with cargo-dist (executable-zips, installers)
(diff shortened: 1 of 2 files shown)


Gankra:
For reference I checked the impl worked by generating this test release of our dummy project:

https://github.com/oranda-gallery/cargodisttest/releases/tag/releases%2Fv0.3.63

(Hand wrote a draft with the right tag set, then pushed a commit with a matching tag.)

MarinPostma (approved):
Thanks @Gankra 🙏

PR #635, opened 2023-08-29
rework namespace creation API

MarinPostma:
This PR makes changes to the namespace creation API. Namely, it:

- the `create-with-dump` route is dropped
- the `create` route takes a payload:
```json
{
    "dump_url": string | null,
}

```

the `dump_url` is either an HTTP url pointing to the dump to be loaded, or, a `file` url, with an absolute path to the dump file.

### DIFF
--- a/sqld/src/admin_api.rs
+++ b/sqld/src/admin_api.rs
@@ -1,13 +1,17 @@
 use anyhow::Context as _;
-use axum::extract::{BodyStream, Path, State};
+use axum::extract::{Path, State};
 use axum::Json;
+use futures::future::OptionFuture;
 use futures::TryStreamExt;
 use serde::Deserialize;
 use std::sync::Arc;
 use std::{io::ErrorKind, net::SocketAddr};
+use tokio_util::io::ReaderStream;
+use url::Url;
 
 use crate::connection::config::{DatabaseConfig, DatabaseConfigStore};
-use crate::namespace::{MakeNamespace, NamespaceStore};
+use crate::error::LoadDumpError;
+use crate::namespace::{DumpStream, MakeNamespace, NamespaceStore};
 
 struct AppState<F: MakeNamespace> {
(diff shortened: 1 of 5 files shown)


sqld/src/admin_api.rs
```diff
@@ -85,23 +90,52 @@ async fn handle_post_block<F: MakeNamespace>(
...
     State(app_state): State<Arc<AppState<F>>>,
     Path(namespace): Path<String>,
-    body: BodyStream,
+    Json(req): Json<CreateNamespaceReq>,
 ) -> Result<(), crate::error::Error> {
-    let dump = Box::new(body.map_err(|e| std::io::Error::new(ErrorKind::Other, e)));
+    let maybe_dump: Option<DumpStream> =
+        OptionFuture::from(req.dump_url.as_ref().map(dump_stream_from_url))
+            .await
+            .transpose()?;
```
LucioFranco, line 101: Can't this all be done with regular rust? Why do we need OptionFuture? In general I would try to avoid using anything from `futures`.

sqld/src/admin_api.rs
```diff
@@ -85,23 +90,52 @@ async fn handle_post_block<F: MakeNamespace>(
...
+
+            if !path.try_exists()? {
+                return Err(LoadDumpError::DumpFileDoesntExist);
+            }
+
+            let f = tokio::fs::File::open(path).await?;
+
+            Ok(Box::new(ReaderStream::new(f)))
+        }
+        _ => todo!("unsupported url scheme"),
```
LucioFranco, line 139: Should this respond with an error instead of a todo panic?

LucioFranco (changes requested):

sqld/src/admin_api.rs
MarinPostma: oops yes

PR #638, opened 2023-08-30
Bottomless: restore from chain of generations

Horusiath:
This PR enables bottomless to perform restore from a generation that may not have a database snapshot included. In general this is realised by introducing a new kind of an object (`.dep`, similar to other metadata objects like `.changecounter` used by snapshot backup and `.meta` used by WAL backup) which stores the dependency between current generation and its predecessor.

This allows us to create a chain of generations that may be necessary (in LIFO order) to be visited in order to restore database state. The algorithm is basically:

1. Try to restore state from a snapshot in a generation we were asked for.
    - If restore from snapshot was ok, proceed to pt.3.
2. If snapshot was not found check the .dep and try to get parent generation.
    - If parent generation was not found (current generation is first one), proceed to pt.3.
    - If parent was found try to stash current generation and try pt.1 for parent. (*this works recursively up to 100 generations deep*).
3. Picking stashed generations one by one, try to find WAL log segments and apply them one by one. When reached the end of a current generation, pick the next stashed one. Rinse and repeat.

I've introduced the next step in bottomless test that verifies this behaviour.

This PR doesn't interfere with snapshots. They are still created on every new generation. It just allows us to fallback to previous generations in case when snapshot is not there - which will be useful once we want to postpone snapshot creation per every generation.


bottomless/src/replicator.rs
```diff
@@ -801,12 +872,12 @@ impl Replicator {
         if let Some(tombstone) = self.get_tombstone().await? {
             if let Some(timestamp) = Self::generation_to_timestamp(&generation) {
                 if tombstone.timestamp() as u64 >= timestamp.to_unix().0 {
-                    tracing::error!(
-                        "Restoration failed. Database '{}' has been tombstoned at {}.",
+                    bail!(
```
Horusiath, line 875: Question to @MarinPostma : this is when we try to restore db when ie. namespace was deleted. How should we react? 

1. Ignore restore with just warning (current status quo)?
2. Return from restore with error (this PR)?

psarna (approved):
Looks good, and just to double-check that I understand everything: we now simply produce more metadata to be able to restore from previous generations in case some snapshot files are missing, and that's it, right? So we have better chance of restoring successfully, and potentially in the future we can optimize by skipping some snapshot files, but it's not the case in this series (which is good, we should favor consistency over storage for now)

Horusiath:
@psarna it's exactly as you've explained.

psarna (approved):

PR #645, opened 2023-08-31
sqld/rpc: Move `Proxy` namespace to metadata

LucioFranco:


sqld/src/connection/write_proxy.rs
```diff
@@ -205,11 +207,12 @@ impl WriteProxyConnection {
         let mut client = self.write_proxy.clone();
 
         let mut req = Request::new(crate::rpc::proxy::rpc::ProgramReq {
-            namespace: self.namespace.clone(),
             client_id: self.client_id.to_string(),
             pgm: Some(pgm.into()),
         });
 
+        let namespace = AsciiMetadataValue::try_from(self.namespace.clone())?;
```
MarinPostma, line 214: Let's encode the header as plain binary data and avoid that error altogether. We don't enforce namespaces to be ASCIII anywhere but here.

MarinPostma: When we later try to parse it, if it is invalid, we can return an `InvalidNamespaceError` which is more useful, wdyt?

LucioFranco: I just made the change to binary, we can revisit later.

MarinPostma (approved):

PR #649, opened 2023-09-04
bottomless: prefix AWS env vars in use with LIBSQL_BOTTOMLESS_

Horusiath:
Currently bottomless is using default AWS SDK env variables. This may cause clashes with other modules that use different credentials and cause issues in the future. For this reason this PR applies the convention (already in use in other bottomless-related env vars) to prefix AWS credentials with `LIBSQL_BOTTOMLESS_`:

- `AWS_DEFAULT_REGION` &rarr; `LIBSQL_BOTTOMLESS_AWS_DEFAULT_REGION`
- `AWS_ACCESS_KEY_ID` &rarr; `LIBSQL_BOTTOMLESS_AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY` &rarr; `LIBSQL_BOTTOMLESS_AWS_SECRET_ACCESS_KEY`

### DIFF
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -6,6 +6,7 @@ use crate::wal::WalFileReader;
 use anyhow::anyhow;
 use arc_swap::ArcSwap;
 use async_compression::tokio::write::GzipEncoder;
+use aws_sdk_s3::config::{Credentials, Region};
 use aws_sdk_s3::error::SdkError;
 use aws_sdk_s3::operation::get_object::builders::GetObjectFluentBuilder;
 use aws_sdk_s3::operation::get_object::GetObjectError;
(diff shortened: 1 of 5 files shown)


bottomless/src/replicator.rs
```diff
@@ -127,6 +151,15 @@ impl Options {
...
         }
+        if let Ok(access_key_id) = std::env::var("LIBSQL_BOTTOMLESS_AWS_ACCESS_KEY_ID") {
+            options.access_key_id = Some(access_key_id);
+        }
+        if let Ok(secret_access_key) = std::env::var("LIBSQL_BOTTOMLESS_AWS_SECRET_ACCESS_KEY") {
+            options.secret_access_key = Some(secret_access_key);
+        }
+        if let Ok(region) = std::env::var("LIBSQL_BOTTOMLESS_AWS_DEFAULT_REGION") {
+            options.region = Some(region);
+        }
```
haaawk, line 162: Why not just:
```
options.access_key_id = Some(std::env::var("LIBSQL_BOTTOMLESS_AWS_ACCESS_KEY_ID")?);
options. secret_access_key = Some(std::env::var("LIBSQL_BOTTOMLESS_AWS_SECRET_ACCESS_KEY")?);
options.region = Some(std::env::var("LIBSQL_BOTTOMLESS_AWS_DEFAULT_REGION")?);
```

Horusiath: Technically it's overriding the default settings, but only if env vars were set. Atm. we don't have any sensible defaults, but this may change.

haaawk (approved):

PR #652, opened 2023-09-05
bottomless: append namespace to db_id

Horusiath:
Instead of replacing `LIBSQL_BOTTOMLESS_DATABASE_ID` assigned by the platform with namespace, we just prepend it with namespace data.

### DIFF
--- a/sqld/src/namespace.rs
+++ b/sqld/src/namespace.rs
@@ -381,7 +381,11 @@ pub type DumpStream =
 
 fn make_bottomless_options(options: &Options, name: &Bytes) -> Options {
     let mut options = options.clone();
-    let db_id = format!("ns-{}", std::str::from_utf8(name).unwrap());
+    let db_id = format!(
+        "ns-{}:{}",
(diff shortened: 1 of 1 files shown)


sqld/src/namespace.rs
```diff
@@ -381,7 +381,11 @@ pub type DumpStream =
 
 fn make_bottomless_options(options: &Options, name: &Bytes) -> Options {
     let mut options = options.clone();
-    let db_id = format!("ns-{}", std::str::from_utf8(name).unwrap());
+    let db_id = format!(
+        "ns-{}:{}",
+        std::str::from_utf8(name).unwrap(),
+        options.db_id.unwrap_or_default()
```
MarinPostma, line 387: this way we can prefix-search all namespaces in a scope
```suggestion
        options.db_id.unwrap_or_default()
        std::str::from_utf8(name).unwrap(),
```

Horusiath: Changed

MarinPostma (approved):

psarna (approved):

PR #653, opened 2023-09-05
check for wal presence before checkpointing

MarinPostma:
This PR adds a check to the logger recovery `checkpoint_db`, that checks if the WAL is present, and if there is anything to checkpoint before performing the checkpoint. This is because litestream interferes with this process on the platform. Instead it is expected that the platform checkpoints the database file before launching sqld.


Horusiath (approved):
It looks like it shouldn't interfere with bottomless behaviour.

PR #656, opened 2023-09-06
fix bad naming for namespace header key

MarinPostma:
Tonic panics if the key for a binary header is not formatted correctly. Tonic doc could be clearer as to what is not a proper formatting, so I had to ding into tonic source to figure it out. And it turns out keys for binary metadata must be suffixed with `-bin`. Now we know :smile:


MarinPostma:
self-merging so we can move on, but FYI @LucioFranco

PR #659, opened 2023-09-06
fix proxy bug

MarinPostma:
When proxying, we always disabled namespaces, so the replica always received an error because the namespace didn't exist.


MarinPostma:
@athoscouto needs that asap, merging

PR #663, opened 2023-09-07
Expose dump endpoint

neubaner:
Expose an endpoint to dump the database content.


MarinPostma (approved):
Looks good! Thanks again! 🙏

PR #672, opened 2023-09-14
Run publish workflow on PRs that target main

athoscouto:


MarinPostma:
I am a bit concerned that:
1) those images are going to override the real release images
2) we are going to flood the registry

Could we use a separate registry for those images, and have them being cleaned up automatically?

MarinPostma:
here is my proposal, actually: let's duplicate this workflow, and have it publish an image with a different name, like `sqld-devel`, and tag that with the PR number, wdyt @athoscouto ?

athoscouto:
Your proposal sounds good and your concerns are valid. I'll close this then.

PR #677, opened 2023-09-15
fix rpc http connector

MarinPostma:
The tls was misconfigured for the server after the networking refactor.

I needed to patch `hyper_rustls` to do that, so I have opened a PR there: https://github.com/rustls/hyper-rustls/pull/221

### DIFF
--- a/sqld/src/rpc/mod.rs
+++ b/sqld/src/rpc/mod.rs
@@ -1,5 +1,10 @@
+use std::sync::Arc;
+
 use anyhow::Context;
 use bytes::Bytes;
+use hyper_rustls::TlsAcceptor;
+use rustls::server::AllowAnyAuthenticatedClient;
+use rustls::RootCertStore;
 use tonic::Status;
 use tower::util::option_layer;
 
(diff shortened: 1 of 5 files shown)


sqld/src/rpc/mod.rs
```diff
@@ -39,33 +44,62 @@ pub async fn run_rpc_server<A: crate::net::Accept>(
...
+            .layer(&option_layer(idle_shutdown_layer))
+            .add_service(ProxyServer::new(proxy_service))
+            .add_service(ReplicationLogServer::new(logger_service))
+            .into_router();
+
+        let h2c = crate::h2c::H2cMaker::new(router);
+        hyper::server::Server::builder(acceptor)
+            .serve(h2c)
+            .await
+            .context("http server")?;
```
LucioFranco, line 89: Is this context gonna hide the error lol? Same with the one below?

LucioFranco (approved):
I left one comment I think we should address then we can merge

sqld/src/rpc/mod.rs
MarinPostma: it's ok, this one is a top-level error, it will show up on the console

PR #688, opened 2023-09-19
Wait for replication index on replica connections

neubaner:
This allows users to wait for writes based on the `replication_index` on replica connections.


MarinPostma (approved):
Thanks for the PR! Sorry for taking so long to review it

PR #690, opened 2023-09-20
Fork with point in time recovery

Horusiath:
This PR basically extends fork API to provide point in time recovery (via optional `{"timestamp":"2023-09-20T10:35:00"}` JSON request body).

### DIFF
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -147,101 +147,66 @@ impl Options {
     }
 
     pub fn from_env() -> Result<Self> {
-        let mut options = Self::default();
-        if let Ok(key) = std::env::var("LIBSQL_BOTTOMLESS_ENDPOINT") {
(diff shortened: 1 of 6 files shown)


bottomless/src/replicator.rs
```diff
@@ -276,13 +241,10 @@ impl Replicator {
         }
 
         let db_path = db_path.into();
-        let db_name = {
-            let db_id = options.db_id.unwrap_or_default();
-            let name = match db_path.find('/') {
-                Some(index) => &db_path[..index],
-                None => &db_path,
-            };
-            db_id + ":" + name
```
Horusiath, line 285: I removed this part - it's a legacy thing that binds part of the database path into a S3 bucket keyspace. It's not necessary as an unique db_id is always provided by sqld, and it would make arbitrary restoration harder.

MarinPostma (approved):

PR #713, opened 2023-09-29
Remove `context` calls for create database

LucioFranco:
Remove the context calls that end up swallowing the real error and make it hard to debug why we couldn't create a database.


psarna (approved):

MarinPostma:
@LucioFranco can you fix clippy when you have time? 🙏

psarna (approved):

PR #727, opened 2023-10-04
changed the path where bottomless is storing gzipped db snapshot

Horusiath:
Changing bottomless snapshot to store db.gz file in the same directory as main db_path.

### DIFF
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -18,7 +18,7 @@ use bytes::{Buf, Bytes};
 use chrono::{NaiveDateTime, TimeZone, Utc};
 use std::io::SeekFrom;
 use std::ops::Deref;
-use std::path::Path;
+use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


bottomless/src/replicator.rs
```diff
@@ -753,14 +767,15 @@ impl Replicator {
...
         let handle = tokio::spawn(async move {
             tracing::trace!("Start snapshotting generation {}", generation);
             let start = Instant::now();
-            let body = match Self::maybe_compress_main_db_file(db_file, compression).await {
+            let body = match Self::maybe_compress_main_db_file(&db_path, compression).await {
                 Ok(file) => file,
                 Err(e) => {
                     tracing::error!(
-                        "Failed to compress db file (generation {}): {}",
+                        "Failed to compress db file (generation {}): {:?}",
```
Horusiath, line 778: Also changed error printout to Debug format - hopefully it will be more informative than `IO error` we got so far.

psarna (approved):

PR #745, opened 2023-10-06
Document `--enable-namespaces` flag under Multitenancy section

avinassh:


avinassh:
I will update this PR with more documentation issues, converting to draft now

PR #751, opened 2023-10-09
do `PRAGMA integrity_check` after db restore

avinassh:
This is similar to `verify` command (#744), after the DB is restored it will do an integrity check. Unlike verify, if the pragma check fails, this keeps the db file around so that we can inspect/analyse it

### DIFF
--- a/bottomless-cli/src/main.rs
+++ b/bottomless-cli/src/main.rs
@@ -2,6 +2,7 @@ use anyhow::Result;
 use aws_sdk_s3::Client;
 use chrono::NaiveDateTime;
 use clap::{Parser, Subcommand};
+use std::path::PathBuf;
 
 mod replicator_extras;
 use crate::replicator_extras::detect_db;
@@ -168,6 +169,12 @@ async fn run() -> Result<()> {
         } => {
             tokio::fs::create_dir_all(&database_dir).await?;
             client.restore(generation, utc_time).await?;
+            let db_path = PathBuf::from(&database);
+            if let Err(e) = verify_db(&db_path) {
+                println!("Verification failed: {e}");
+                std::process::exit(1)
+            }
+            println!("Verification: ok");
         }
         Commands::Verify {
             generation,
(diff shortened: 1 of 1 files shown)


bottomless-cli/src/main.rs
```diff
@@ -168,6 +169,12 @@ async fn run() -> Result<()> {
         } => {
             tokio::fs::create_dir_all(&database_dir).await?;
             client.restore(generation, utc_time).await?;
+            let db_path = PathBuf::from(&database);
+            let result = verify_db(&db_path).unwrap_or_else(|e| e.to_string());
+            println!("Verification: {result}");
```
MarinPostma, line 174: oops :)

bottomless-cli/src/main.rs
```diff
@@ -205,6 +209,14 @@ async fn run() -> Result<()> {
     Ok(())
 }
 
+fn verify_db(path: &PathBuf) -> Result<String> {
```
MarinPostma, line 212: you should verify the result returned by the pragma here, and return a `Result<()>` instead

MarinPostma (changes requested):

MarinPostma:
One small thing, and also you need to run rust-fmt :)

bottomless-cli/src/main.rs
avinassh: oh right, thats a good idea :) 

I have updated it. Is there any better alternative to `anyhow`, since we plan to remove it

MarinPostma: I think it can stay inside of bottomless bin, but the alternative to anyhow is to have our own Error enum, and categorize everything manually :)

bottomless-cli/src/main.rs
```diff
@@ -205,6 +210,18 @@ async fn run() -> Result<()> {
...
 
+fn verify_db(path: &PathBuf) -> Result<()> {
+    let conn = rusqlite::Connection::open(path)?;
+    let mut stmt = conn.prepare("PRAGMA integrity_check")?;
+    let mut rows = stmt.query(())?;
+    let result: String = rows.next()?.unwrap().get(0)?;
+    if result == "ok" {
+        Ok(())
+    } else {
+        Err(anyhow::anyhow!(result.to_string()))
```
MarinPostma, line 221:
```suggestion
        anyhow::bail!("{result}")
```

MarinPostma: for a followup

MarinPostma (approved):

PR #761, opened 2023-10-11
Make cargo-chef effective for CI builds

athoscouto:
This PR has two changes that should make `cargo-chef` effective for CI docker builds.

- Use another image (`athosturso/sqld-builder`, built from `Dockerfile.builder`) as a base for both `cargo-chef` steps.
- Enable registry caching for Docker layers, allowing Docker layers to be reused between runs.

The first step is necessary because we can't reuse cached layers if the previous layers have changed.
And layers that run `apt update`, `rustup update` and `cargo install` can invalidate those caches.
The solution is to create an external image that has a separate lifecycle and does those unreproducible operations.

If this is accepted, we should create an `sqld-builder` package to store images for `Dockerfile.builder`.

The second step generates a tag inside the package to store the build cache (see [this](https://github.com/libsql/sqld/pkgs/container/sqld-devel/136494657?tag=buildcache) for example).
We may choose to have a different package to store only build caches as well.

---

When no dependencies are changed and `cargo-chef` works correctly, it takes [4 minutes](https://github.com/libsql/sqld/actions/runs/6486191875/job/17613868839) to build the Docker image, instead of [20](https://github.com/libsql/sqld/actions/runs/6485615033).

### DIFF
--- a/.github/workflows/pr-images.yml
+++ b/.github/workflows/pr-images.yml
@@ -22,6 +22,9 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+  
       - name: Log in to the Container registry
         uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
         with:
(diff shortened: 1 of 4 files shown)


athoscouto:
My understanding of how Docker cache works was incorrect, we don't need the extra image.
Closing this in favor of #763.

PR #773, opened 2023-10-13
error: fix typo and normalize messages with a Title starting word

ultrabug:
The maniac in me hates errors, and hates them even more when they are formatted inconsistently :stuck_out_tongue:


MarinPostma:
thanks

